PR #3, opened 2015-06-19
WIP connect request for existing nodes

benjaminbollen:
This is an important security flaw in the current implementation of address relocation!  This needs a urgent solution, but the solution is achievable with the tools available.


dirvine:
This is quite a long way from being complete or considered at the moment. I presume this is a placeholder at the moment

PR #10, opened 2015-06-24
RFC - Use public keys only as identifier of various clients and actions

dirvine:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/10)

### DIFF
--- a/proposed/0000-Use-public-key-for-id-on-all-messages.md
+++ b/proposed/0000-Use-public-key-for-id-on-all-messages.md
@@ -1,0 +1,60 @@
+- Feature Name: Use public key alone to authorise actions 
+- Type Enhancement
+- Related components routing, types, client, vaults
+- Start Date: 23-06-2015
+- RFC PR: 
+- Issue number: 
+
+# Summary
+
+At the moment client Id packets are create and a hash of these packets used as identity. 
+This involves storing such packets prior to the client interacting with the network. 
+As such this requires an initial `unauthorised put` as the client is not known to the 
+network and cannot be recognised without this. It also means there are many lookups
+to link the hash of the id packet with the client taking the action. 
+These lookups are costly it is proposed here that they are not required.  
+
+# Motivation
+
+Storing Id packets is a strain on the network. Lookups are a strain and will slow down
+network actions. Id packets themselves may be a security issue (can they be hacked), although
+highly unlikely, it is impossible to hack if they do not exist! 
+
+This mechanism does not reduce the total number of clients or types of clients as they all exist 
+in an address space of 2^256 and app developers can use multiple Id's in very clever ways, such 
+as the SAFE network does to distinguish public id's and private id's. Now these and more are possible.
+Id's could be created to share data types (as co-owners) and allow shares access and write capability
+between applications and groups. 
+
+# Detailed design
+
+As clients `Put` data or messages on the network (i.e. ask to create) they send such requests via the 
+ClientManager type persona (MaidManager in the case of data). These requests are signed and include the
+clients `PublicKey`. This public key can be allocated, or matched to an existing,to a client account. 
+The ClientManager may then take any action, such as authorising the `Put`, taking a balance etc. and all against the key. 
+
+If a client wishes to amend any data (structured data) then this is already signed by that client and contains the 
+`PublicKey` The client will sign the request to alter the data (by overwriting with a new valid one) and again
+the public key of the signed message is included in the message, confirming what key made the request. This key
+can then be confirmed to be one of the owners of the `StructuredData` element and if multiple owners
+then the signatures of the data element being uploaded is checked against the list of signatures of the new chunk
+using the owners keys of the last chunk. 
+
+Clients may identify themselves using the public key as their node address. Signing the request to join the network 
+at that address. This is a different size key from routing nodes though, which is a benefit as it distinguishes these
+connections more clearly. This does require a change to the routing API to accommodate 32 byte node identities.
+
+Clients themselves no longer will require to store Id's packets on the network (although they can via StructuredData).
+`PublicId` types are no longer required in the maidsafe_types library (maidsafe_client exclusive now).
+
+# Drawbacks
+
+To be identified
+
+# Alternatives
+
+Alternatively the status quo could be left in place
+
+# Unresolved questions
+
+To be resolved during evaluation.
--- a/proposed/0000-Use-public-key-for-id-on-all-messages.md
+++ b/proposed/0000-Use-public-key-for-id-on-all-messages.md
@@ -7,12 +7,24 @@
 
 # Summary
 
-At the moment client Id packets are create and a hash of these packets used as identity. 
-This involves storing such packets prior to the client interacting with the network. 
+At the moment client Id packets are created and a hash of these packets used as identity.
+This is stored on th network as a `key` (ie the hash) / `value` (the public key) pair.
+These packets have to be available prior to the client storing data onto the network. 
+
+As the network must confirm a client has paid the network, either by providing resource
+or via a network token (i.e. safecoin). Then it needs to look up the clients public key 
+and confirm the signatuire of requests come from that client. This is done by querying 
+the client ID (Hash), then looking up the ID packet and downloading it to get the public key.
+
 As such this requires an initial `unauthorised put` as the client is not known to the 
-network and cannot be recognised without this. It also means there are many lookups
-to link the hash of the id packet with the client taking the action. 
-These lookups are costly it is proposed here that they are not required.  
+network and cannot be recognised without this. This means essentially the network has to 
+allow unliited `Put` of client Id packets, thereby exposing a risk of wasted network usage.
+
+This proposal removes all of this indirection and instead allows the client to be recodnised
+by the public key included in messages or data types. such messages and data types will be 
+signed by the `secret key` that is paired with this public key. As only the owner should have
+access to this `secret key` then it can be assumed the message or request to mutate data is 
+indeed valid in a cryptographically secure manner.
 
 # Motivation
 
--- a/proposed/0000-Use-public-key-for-id-on-all-messages.md
+++ b/proposed/0000-Use-public-key-for-id-on-all-messages.md
@@ -22,7 +22,7 @@ highly unlikely, it is impossible to hack if they do not exist!
 
 This mechanism does not reduce the total number of clients or types of clients as they all exist 
 in an address space of 2^256 and app developers can use multiple Id's in very clever ways, such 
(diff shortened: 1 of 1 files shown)


proposed/0000-Use-public-key-for-id-on-all-messages.md
```diff
@@ -0,0 +1,60 @@
+- Feature Name: Use public key alone to authorise actions
```
benjaminbollen: only or alone?

proposed/0000-Use-public-key-for-id-on-all-messages.md
```diff
@@ -0,0 +1,60 @@
...
+
+# Motivation
+
+Storing Id packets is a strain on the network. Lookups are a strain and will slow down
+network actions. Id packets themselves may be a security issue (can they be hacked), although
+highly unlikely, it is impossible to hack if they do not exist! 
+
+This mechanism does not reduce the total number of clients or types of clients as they all exist 
+in an address space of 2^256 and app developers can use multiple Id's in very clever ways, such 
+as the SAFE network does to distinguish public id's and private id's. Now these and more are possible.
```
mmoadeli: as private ids are not stored on network, how does safe network need to distinguish between public and private ids?

proposed/0000-Use-public-key-for-id-on-all-messages.md
```diff
@@ -0,0 +1,60 @@
...
+These lookups are costly it is proposed here that they are not required.  
+
+# Motivation
+
+Storing Id packets is a strain on the network. Lookups are a strain and will slow down
+network actions. Id packets themselves may be a security issue (can they be hacked), although
+highly unlikely, it is impossible to hack if they do not exist! 
+
+This mechanism does not reduce the total number of clients or types of clients as they all exist 
+in an address space of 2^256 and app developers can use multiple Id's in very clever ways, such
```
mmoadeli: why 2^256? why not 2^512?

proposed/0000-Use-public-key-for-id-on-all-messages.md
```diff
@@ -0,0 +1,60 @@
...
+# Detailed design
+
+As clients `Put` data or messages on the network (i.e. ask to create) they send such requests via the 
+ClientManager type persona (MaidManager in the case of data). These requests are signed and include the
+clients `PublicKey`. This public key can be allocated, or matched to an existing,to a client account. 
+The ClientManager may then take any action, such as authorising the `Put`, taking a balance etc. and all against the key. 
+
+If a client wishes to amend any data (structured data) then this is already signed by that client and contains the 
+`PublicKey` The client will sign the request to alter the data (by overwriting with a new valid one) and again
+the public key of the signed message is included in the message, confirming what key made the request. This key
```
mmoadeli: "public key of the signed message is included in the message"?

proposed/0000-Use-public-key-for-id-on-all-messages.md
```diff
@@ -0,0 +1,60 @@
...
+can then be confirmed to be one of the owners of the `StructuredData` element and if multiple owners
+then the signatures of the data element being uploaded is checked against the list of signatures of the new chunk
+using the owners keys of the last chunk. 
+
+Clients may identify themselves using the public key as their node address. Signing the request to join the network
```
mmoadeli: Clients may identify themselves using the <public key> as their node address ? 
"hash" of public key?

proposed/0000-Use-public-key-for-id-on-all-messages.md
```diff
@@ -0,0 +1,60 @@
+- Feature Name: Use public key alone to authorise actions 
+- Type Enhancement
+- Related components routing, types, client, vaults
```
mmoadeli: maidsafe_types, maidsafe_client, maidsafe_vault

proposed/0000-Use-public-key-for-id-on-all-messages.md
dirvine: A public Id would be publicly known id (i.e. David or MPID) private ID would be what we seen as `Maid`. The network does not need to distinguish these as they are the responsibility of the client apps. i.e. a client app should use a public known name for messaging/contacts for insance

proposed/0000-Use-public-key-for-id-on-all-messages.md
dirvine: limited by public key size (32 byte), so it's a physical limit of the encryption algorithm. We could increase that key size and eventually go to 2^512 but for time being it's computationally to expensive to do so.

proposed/0000-Use-public-key-for-id-on-all-messages.md
dirvine: Perhaps, or maybe "The public key associated with the signed message"

proposed/0000-Use-public-key-for-id-on-all-messages.md
dirvine: I am not sure clients should not actually use a 32 byte node address. It is more efficient and also means there is a real distinction as well as not polluting the 64 byte space.

proposed/0000-Use-public-key-for-id-on-all-messages.md
dirvine: Cheers, I will alter this.

proposed/0000-Use-public-key-for-id-on-all-messages.md
mmoadeli: It was exactly the point I raised. That safe network does not need to distinguish.

dirvine: Are you agreeing with the rfc and the comment I made on it? confused! Ah I think I See what you are saying, the work `not` is missing :-)

mmoadeli: "not" is not added to the updated file as I see.

proposed/0000-Use-public-key-for-id-on-all-messages.md
mmoadeli: The only point I am missing is that you mentioned "Clients may identify themselves using the public key as their node address"
and above you mention "I am not sure clients should not actually use a 32 byte node address"

what is the  relationship between this 32 byte node address & public key?

dirvine: e25519 public key size == 32Bytes

mmoadeli: Thanks

mmoadeli: I would mention it explicitly earlier in ths document . If no where else, in our documentation it is mentioned.

proposed/0000-Use-public-key-for-id-on-all-messages.md
```diff
@@ -0,0 +1,60 @@
...
+ClientManager type persona (MaidManager in the case of data). These requests are signed and include the
+clients `PublicKey`. This public key can be allocated, or matched to an existing,to a client account. 
+The ClientManager may then take any action, such as authorising the `Put`, taking a balance etc. and all against the key. 
+
+If a client wishes to amend any data (structured data) then this is already signed by that client and contains the 
+`PublicKey` The client will sign the request to alter the data (by overwriting with a new valid one) and again
+the public key of the signed message is included in the message, confirming what key made the request. This key
+can then be confirmed to be one of the owners of the `StructuredData` element and if multiple owners
+then the signatures of the data element being uploaded is checked against the list of signatures of the new chunk
+using the owners keys of the last chunk.
```
benjaminbollen: This also puts responsability on the client/app for multisig to negotiate amongst the owners to agree on a new version and delegate one client to collect all signatures and execute the put on behalf of all;

alternatively we can also put such behaviour in "Sentinel-like accumulator", where clients would have to coordinate their put; this allows a double route though (as it would not exclude the above), and maybe thats not a desirable effect; also Sentinel should not be involved here as the core idea is for the put request to be self-verifiying

proposed/0000-Use-public-key-for-id-on-all-messages.md
```diff
@@ -0,0 +1,60 @@
...
+If a client wishes to amend any data (structured data) then this is already signed by that client and contains the 
+`PublicKey` The client will sign the request to alter the data (by overwriting with a new valid one) and again
+the public key of the signed message is included in the message, confirming what key made the request. This key
+can then be confirmed to be one of the owners of the `StructuredData` element and if multiple owners
+then the signatures of the data element being uploaded is checked against the list of signatures of the new chunk
+using the owners keys of the last chunk. 
+
+Clients may identify themselves using the public key as their node address. Signing the request to join the network 
+at that address. This is a different size key from routing nodes though, which is a benefit as it distinguishes these
+connections more clearly. This does require a change to the routing API to accommodate 32 byte node identities.
```
benjaminbollen: would there still be a need for a client to directly connect to its client manager group?  Can we not do everything over an arbitrary relay node?  The RelayNode would then map to a 32bit "name" - but the relay node should not be too concerned with verifying this name as it only matters to the content and groups would decide on that.  Does that make sense?

dirvine: The only time would be when we want push data from the network (so messaging for instance, although a register current location can be used instead). To prevent spam a relay could cryptographically check the nodeid, but this proposal does not prohibit that in future.

proposed/0000-Use-public-key-for-id-on-all-messages.md
dirvine: Yes it does. That was the intent really, that on successful gathering of the signatures a single person pushes the button (so to speak).  I think this is suited to client side as you can then create out of band processes like board level votes, parliament voting systems etc. where the system recognises there are enough sigs and sends the update.

proposed/0000-Use-public-key-for-id-on-all-messages.md
```diff
@@ -0,0 +1,60 @@
...
+
+# Summary
+
+At the moment client Id packets are create and a hash of these packets used as identity. 
+This involves storing such packets prior to the client interacting with the network. 
+As such this requires an initial `unauthorised put` as the client is not known to the 
+network and cannot be recognised without this. It also means there are many lookups
+to link the hash of the id packet with the client taking the action. 
+These lookups are costly it is proposed here that they are not required.  
+
```
inetic: I think, it should be elaborated on what is the reasoning behind the current mechanism and how the mechanism works (either here or providing a link to documentation). Also a formal definition of `action` would be handy somewhere. Otherwise I think it's not a RFC, but just a memo for those who already know the context.

dirvine:

PR #13, opened 2015-06-24
RFC - Farm Attempt

dirvine:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/13)

### DIFF
--- a/proposed/0000-Farm-attempt.md
+++ b/proposed/0000-Farm-attempt.md
@@ -21,7 +21,7 @@ This implementation will be solely in the vault library. This will include sever
 
 ON receipt of a `Get` request for `ImmutableData` the `DataManager` calculates the SHA512 Hash of the `name()` of the `ImmutableData` concatenated with the`PmidNode`. This result is then modulo divided by the farming rate (FR). The rate attempts should divide the award between 4 distinct groups. (for clarity, this algorithm is applied to all `PmidNodes` that hold this data element in the current group).
 
-##Rmte attempts 
+##Rate attempts 
 
 1. PmidNodes -> tested at 100% of FR [i.e. Farming rate * 1]
 2. App Developer -> tested at 10% of FR [i.e. Farming rate * 1.1] 
--- a/proposed/0000-Farm-attempt.md
+++ b/proposed/0000-Farm-attempt.md
@@ -1,0 +1,56 @@
+- Feature Name:Farm attempt 
+- Type new feature
+- Related components maidsafe_vault
+- Start Date: 24-06-2015 
+- RFC PR:
+- Issue number: 
+
+# Summary
+
+This proposal outlines an initial design to test the efficiency and security of farming attempts.  This involves making use of the Farming Rate (FR) which is the rate at which farming attempts are made.  It is assumed the farming rate will include payments to care development, content producers and most significantly farmers (or providers of the network resource). 
+
+# Motivation
+
+To ensure network stability and resource availability a reward mechanism is required to ensure the use of resource is balanced with the supply of resource. 
+
+# Scope of work
+
+This implementation will be solely in the vault library. This will include several steps. The client lib will require to be aware of the actual safecoin type (again `StructuredData`). 
+
+# Detailed design
+
+ON receipt of a `Get` request for `ImmutableData` the `DataManager` calculates the SHA512 Hash of the `name()` of the `ImmutableData` concatenated with the
+`PmidHolder`. This result is then modulo divided by the farming rate (FR). The rate attempts should divide the award between 4 distinct groups. 
+
+##Rmte attempts 
+
+1. PmidNodes -> tested at 100% of FR [i.e. Farming rate * 1]
+2. App Developer -> tested at 10% of FR [i.e. Farming rate * 1.1] 
+3. Publisher -> tested at 10% of FR [i.e. Farming rate * 1.1]
+4. Core development -> tested at 5% of FR [i.e. Farming rate * 1.2]
+
+##Payment address 
+
+1. PmidNodes -> Registers an Optional (if set by user) wallet address on any store of data to it
+2. App Developer -> App developers will include wallet address in any `Get` request
+3. Publisher -> As data is stored a wallet address of the owner is stored if this is first time seen on the network (stored in DM account for the data element)
+4. Core development -> Initially every node will be aware of a hard coded wallet address for core development. This will likely lead to a multisig wallet.
+
+##Farm process
+
+When the initial farming test above is true (the modulo division ==0 ) then the `DataManagers` send a Farm request message `Post` to the group closest to the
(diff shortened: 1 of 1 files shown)


chrisfoster:
+4. Core development -> tested at 5% of FR [i.e. Farming rate * 1.2]

5% of FR = 1.05 ?

dirvine:
Ah Thanks Chris

proposed/0000-Farm-attempt.md
```diff
@@ -0,0 +1,56 @@
...
+3. Publisher -> As data is stored a wallet address of the owner is stored if this is first time seen on the network (stored in DM account for the data element)
+4. Core development -> Initially every node will be aware of a hard coded wallet address for core development. This will likely lead to a multisig wallet.
+
+##Farm process
+
+When the initial farming test above is true (the modulo division ==0 ) then the `DataManagers` send a Farm request message `Post` to the group closest to the
+combined Hash (i.e. Hash of (`PmidHolder` + `ImmutableData` )). This request includes the original hashes + hash of chunk requested +current farming rate of the group as well as wallet address. This is confirmed at the receiving group and they check for the existence of a safecoin space (ie, there is no safecoin with this name). If this is successful then a further check is made to confirm there is a `DataManager` group (implicit as routing does this). A safecoin is then created and a message sent to the wallet holder of the request.  All these requests must be digitally signed and confirmed at receiving end (implicit in sentnel).
```
chandraprakash: Typo: 
   (i.e. Hash of (`PmidHolder` + `ImmutableData` ))
Order should be same as explained in detail design section.
   `DataManager` calculates the SHA512 Hash of the `name()` of the `ImmutableData` concatenated with the
   `PmidHolder`

proposed/0000-Farm-attempt.md
```diff
@@ -0,0 +1,56 @@
...
+To ensure network stability and resource availability a reward mechanism is required to ensure the use of resource is balanced with the supply of resource. 
+
+# Scope of work
+
+This implementation will be solely in the vault library. This will include several steps. The client lib will require to be aware of the actual safecoin type (again `StructuredData`). 
+
+# Detailed design
+
+ON receipt of a `Get` request for `ImmutableData` the `DataManager` calculates the SHA512 Hash of the `name()` of the `ImmutableData` concatenated with the
+`PmidHolder`. This result is then modulo divided by the farming rate (FR). The rate attempts should divide the award between 4 distinct groups.
```
chandraprakash: I assume `PmidHolder` means one of the PmidNode. 
Currently DataManager Sends on the get request to all the pmid_nodes. 
So, we need to have a deterministic way to get the PmidHolder(s), so that the hash at different group nodes yield same result.

proposed/0000-Farm-attempt.md
```diff
@@ -0,0 +1,56 @@
...
+- RFC PR:
+- Issue number: 
+
+# Summary
+
+This proposal outlines an initial design to test the efficiency and security of farming attempts.  This involves making use of the Farming Rate (FR) which is the rate at which farming attempts are made.  It is assumed the farming rate will include payments to care development, content producers and most significantly farmers (or providers of the network resource). 
+
+# Motivation
+
+To ensure network stability and resource availability a reward mechanism is required to ensure the use of resource is balanced with the supply of resource.
```
mmoadeli: I would change
"To ensure network stability and resource availability" 
to 
"As an incentive to improve network stability and resource availability"

proposed/0000-Farm-attempt.md
dirvine: Yes there will be 2 PmidNodes in primary group, so we test both against this algorithm.

proposed/0000-Farm-attempt.md
dirvine: Noted.

proposed/0000-Farm-attempt.md
chandraprakash: +1 .. keeps it simple.

proposed/0000-Farm-attempt.md
```diff
@@ -0,0 +1,56 @@
...
+##Payment address 
+
+1. PmidNodes -> Registers an Optional (if set by user) wallet address on any store of data to it
+2. App Developer -> App developers will include wallet address in any `Get` request
+3. Publisher -> As data is stored a wallet address of the owner is stored if this is first time seen on the network (stored in DM account for the data element)
+4. Core development -> Initially every node will be aware of a hard coded wallet address for core development. This will likely lead to a multisig wallet.
+
+##Farm process
+
+When the initial farming test above is true (the modulo division ==0 ) then the `DataManagers` send a Farm request message `Post` to the group closest to the
```
mmoadeli: ... message `Post` to the "group" closest to the ....
what persona the "group" has?

mmoadeli:
These RFCs are the most useful entities.

proposed/0000-Farm-attempt.md
dirvine: That will be the `NaeManager` group, probably a specialised DataManager as it handles some additional functionality than just storing a data element (it' checks from authority for these types + may include further checks)

dirvine:
:+1:

proposed/0000-Farm-attempt.md
chandraprakash: Will be nice if you can add a line saying that same algorithm will be applied on individual `PmidNode` storing the data.
Also replace `PmidHolder` to `PmidNode`.

benjaminbollen:
I propose a pay-it-forward deflection to avoid that PmidNodes can harvest their own safecoins.

1. GET for ImmutableData abc -> DM post to H(abc + pmid1 + pmid2) = def a safecoin-challenge;
2. all DM for 'def' forward the challenge to DM-group for the chunk closest to the challenge -- possible challenge splits and does not receive quorum and terminates; say consensus is reached for chunk deg
3. DM at deg post challenge def to PNs storing deg; these PNs respond with full chunk back to DMs, that check correct storage; on successful return of all PNs award award to PNs of deg, not abc

benjaminbollen:
possibly execute step 3. storage verification at BackupImmutable chunks; that way primary chunk holders are needed for deflecting safecoin-challenges; and backup holders are rewarded for redundancy efforts; Sacrificial holders are needed for pushing up the farming rate; that way all 3 levels positively contribute to the chances of farming safecoin;

benjaminbollen:
note: the forward deflection pattern we have already implemented for address relocation

proposed/0000-Farm-attempt.md
```diff
@@ -0,0 +1,55 @@
...
+3. Publisher -> tested at 10% of FR [i.e. Farming rate * 1.1]
+4. Core development -> tested at 5% of FR [i.e. Farming rate * 1.05]
+
+##Payment address 
+
+1. PmidNodes -> Registers an Optional (if set by user) wallet address on any store of data to it
+2. App Developer -> App developers will include wallet address in any `Get` request
+3. Publisher -> As data is stored a wallet address of the owner is stored if this is first time seen on the network (stored in DM account for the data element)
+4. Core development -> Initially every node will be aware of a hard coded wallet address for core development. This will likely lead to a multisig wallet.
+
```
maqi: the above wallet address stuff sounds more like an additional function needs to be added in MM,
or having an additional persona called WalletManager to handle the wallet management

dirvine: Yes, @maqi Good point, I see when safecoin is minted for anyone a network message will go to their wallet. So perhaps wallet manager is something we should consider, but may be handled by messaging?

dirvine:

PR #14, opened 2015-06-25
RFC Balance network resources

dirvine:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/14)

### DIFF
--- a/proposed/0000-balance_network_resources.md
+++ b/proposed/0000-balance_network_resources.md
@@ -1,0 +1,76 @@
+- Feature Name: Balance network resources 
+- Type new feature
+- Related components maidsafe_vault 
+- Start Date: 25-06-2015 
+- RFC PR: 
+- Issue number: 
+
+# Summary
+
+This proposals intent is to provide a balance for Farming Rate verses purchase price of network
+resources that users are charged. The farming rate is calculated by the network in another RFC.
+As the farming rate increases, farmers earn less as more `Get` requests per attempt is required.
+As the farming rate increases then it follows users pay less (as the network is oversupplied) for
+resources. The key issue is to identify the link between these two numbers. It is assumed
+all chunks are 1Mb in this proposition, this is very realistic in terms of calculation and 
+any error favours the user. 
+
+# Motivation
+
+This decentralised autonomous network must encourage resource providers. This is anticipated
+to be similar to skype, bittorrent etc. in that people will provide resources, simply because 
+they are getting something from the network. This proposal enhances safecoin to provide an
+additional encouragement. That incentive is the payment of providers of resources to include
+application development, content availability and importantly resources, in terms of disk space,
+cpu processing, memory, bandwidth etc. in fact everything required to provide data availability
+(and compute) to all users, irrespective of the use or users.
+
+Many confuse any out of network price fluctuation would in fact alter the network balancing process, 
+in fact it is unrelated as the balance simply increases reward when space is needed and reduces when
+space is abundant. There is no relation to the cost of safecoin and network balancing, as long as 
+a safecoin can change hands for a cost then the number of safecoin per unit of fiat is handled by
+the network.
+
+# Detailed design
+
+A very simple approach is to use a fixed approach and measure via testing. It is assumed there will 
+be considerably more reads from the network than stores of data. The cost per MB, electricity, B/W 
+and size of storage are all factors in this algorithm. Therefore testing and sampling of a running 
+network is essential. To begin the simple approach is best. 
+
+`int cost_per_Mb = 1/(fr^(1/5));`  (5th root, integer value)
+
+Where 1 is a 1MB chunk. It is assumed the figure will alter under scrutiny of network testing, but 
+this is assumption (based on usage increase (X10 pa total), price decrease (1/2 pa) etc. over time)
+as well as b/w usage, which widely varies across countries at the moment. 
+
+The FR is the local farming rate known to the `ClientManager` and may slightly vary through time across
+the network. This is expected and as the network grows the spread of cost will equalise as the binary 
+tree balances. 
+
+`ClientManagers` will expect at least a whole safecoin in advance, which is `burned` (i.e. the client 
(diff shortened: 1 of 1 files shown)


benjaminbollen:
A comment to provoke discussion: is this not doubly balancing the network resources?  If farming rate is already adjusted based on network resources; linked with that the rate of earning on Get; should the price for Put not simply be fixed in safecoin?

say - just using 1 as a factor everywhere - Put 1 chunk (1 Mb) = 1 safecoin; then the network at most rewards 1 safecoin for the safecoin-challenge associated to that chunk (the challenge is created by Get requests).  For narrowing network resources, sacrificial Put fails which decreases the Farming rate (making farmers earn more); which after some time pushes down the market-price of safecoin; which makes Put cheaper, helping to fill up the network further and pushing the price of safecoin up.

dirvine:
>should the price for Put not simply be fixed in safecoin?

If the external price of safecoin goes wild then this ties that price to the data store price. The proposal I think balances this so that a payment should (at least the intention is) means users pay considerably less safecoin for storing as external price increases. Farmers get less safecoin as external price increases in this manner as well. This is why there is an apparent double balance.

benjaminbollen:
but if the 'market-price' of safecoin goes wild -either way-  would the 'quid-pro-quo' nature not allow the network to function still; you farm to earn safecoin that you can spend to store a proportional amount of data? regardless of what the outside world considers the fiat price to be?

dirvine:
Yes, well sort of, the perception of paying the equivalent of $XXX per Mb would though make folks value perception see the network as expensive (from users point) and drive people to alternatives I imagine. So the double balance means fokls see the real cost (even in fiat terms) of storage, which should be cheapest price possible as the network is basically acting like a stock marker for resources.

benjaminbollen:
I keep arguing on opposite point mostly to get to the bottom of this --  I would argue that allowing the Put price (MB / safecoin) to depend on the farming rate, will be confusing to users; on day 1 they load their balance by deleting safecoin, and have an account of storing X MB.  After that the local farming rate has changed (without them actively knowing) and the available balance on their account in MB has gone up or down.  I argue that that is confusing, and certainly if the farming rate would be volatile (which it might not be, and should not be) then that might give an unpleasant user experience

benjaminbollen:
or I am maybe wrong in my description here: once the safecoin is converted at the current safecoin <-> MB put price; then their account is fixed in MB. :)

dirvine:
>once the safecoin is converted at the current safecoin <-> MB put price; then their account is fixed in MB. :)

yes

benjaminbollen:
For later considerations: an alternative to linking put price to the farming rate is to consider the local ratio of successful farming challenges over all (local) farming challenges creates.  This could be considered a step closer to the actual safecoins awarded, than the farming rate.  Potentially worth looking at the work of D Gordon on the foraging rate of ants, for mathematical models.

AKlinky:
Another way to look at this is considering the SAFE network as a decentralized autonomous (non-profit) organization. The SAFE network has income (from PUTting clients), expenditures (rewards to farmers), and capital (non-issued SafeCoins). 

At some point in the future no more new SafeCoins can be issued, but SafeCoins will still be recycled. This means for the SAFE network to maintain equilibrium (not "bankrupting") in this late stage, income (over any given period) needs to at least match expenditures. So in general, the network should ask just enough SafeCoin from PUTting clients to be able to accommodate vaults. This way the absolute lowest PUT price will be found, maximizing accessibility for all.

One way to do this is having a target amount of SafeCoins in existence. At the late stage of the network, this could perhaps be 99%. If over 99% of SafeCoins are in existence, PUT prices would be raised, if less than 99%, PUT prices would be lowered. The remaining 1% would function as a buffer to protect the network against sudden fluctuations in demand (amount of PUTS).

If such an algorithm would be used, it would seem logical to me to also have a target amount of SafeCoins in existence before this late stage, right from the start. The target could be derived from any variable or combination of variables the network can autonomously measure or approximate (for example current total network capacity).

As an added bonus, if derived wisely, this usage of a target amount of SafeCoin could also protect against potential malign high inflation rates in the early days of the network, when a lot of new SafeCoins would be issued. I must admit I'm not aware how predictable and balanced the current algorithm of issuance of new SafeCoins is, so maybe malign high inflation is already impossible.

- Seneca

dirvine:
Nice @AKlinky  I think you have put your finger right n the pulse here. The recycle is hugely important and the balance of users payment to resource available is paramount. This RFC assumes resources keep getting cheaper and users require more resources. I think both of these are almost a given now and advances in b/w etc. will make this even more so. 

The network will always want safecoin to survive, but humanity (hopefully) should always want more information and calculation, in this view then the value of the network will hopefully factor highly in peoples minds. I also think that the IOT (Internet Of Things) can take advantage here as well where vaults can fit into very small embedded devices who may only hold a few thousand chunks at a time, but again even these will be faster and hold more data in an exponential growth pattern. 

The hope is that the capacity will have no issue catching up with the demand as long as uploaders pay to ensure the resources do in fact balance. 

This RFC is looking for the ultimate algorithm to balance this in a manner that decays cost of resource over time. We imagine this will alter over time as the network use pattern evolves and morphs, the perfect algorithm will take varying use patterns into account. It's a nice thought experiment for sure, but we will start with measurements and tweaks. Then use the RFC process to tweak further. What's nice is that it's almost impossible for core developers to manipulate this for self gain, it can only be for network gain, therefore everyone should benefit.

dirvine:

proposed/0000-balance_network_resources.md
```diff
@@ -0,0 +1,76 @@
...
+- Start Date: 25-06-2015 
+- RFC PR: 
+- Issue number: 
+
+# Summary
+
+This proposals intent is to provide a balance for Farming Rate verses purchase price of network
+resources that users are charged. The farming rate is calculated by the network in another RFC.
+As the farming rate increases, farmers earn less as more `Get` requests per attempt is required.
+As the farming rate increases then it follows users pay less (as the network is oversupplied) for
```
maqi: shall be "As the framing rate decreases then it follows users pay less" here?

PR #19, opened 2015-06-27
RFC - Name Service

dirvine:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/19)

### DIFF
--- a/proposed/0000-name-service.md
+++ b/proposed/0000-name-service.md
@@ -1,0 +1,67 @@
+- Feature Name: Decentralised Naming System (dns)
+- Type new feature
+- Related components maidsafe_client, maidsafe_vault
+- Start Date: 27-06-2015
+- RFC PR:
+- Issue number: 
+
+# Summary
+
+In the current Internet Domains are referenced via a naming system which comprises an indexing mechanism
+This indexing mechanism is centralised and controlled, with the ability for countries and ISPs to switch it off
+pollute data with advertising data and much worse. The SAFE network on the other had offers a 'free (as in beer)'
+mechanism to locate data linked to a name. This proposal outlines a mechanism to look up data related to any name.
+
+This data includes, long name (could be real name is users wish), web site id, blog id, micro-blog id and may be 
+extended to include additional information as the network develops. 
+
+# Motivation
+
+In networks people expect to be able to lookup services related to names. In the SAFE network this includes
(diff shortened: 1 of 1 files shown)


dirvine:

proposed/0000-name-service.md
```diff
@@ -0,0 +1,67 @@
...
+struct dns {
+long_name: String
+encrytion_key: crypto::encrypt::public_key
+//      service, location
+HashMap<String, Identity>
+}
+
+Initially this `HashMap` will likely contain www, blog and micro-blog. Application developers may add services to enhance 
+users experiences with new applications and functions via this mechanism. 
+
```
ustulation: So just as example: Say I want to create 2 web-sites.
So HashMap entries will be:
`www.website0.com` => `Root Directory ID 0 that contains HOME.html etc files`
`www.website1.com` => `Root Directory ID 1 that contains HOME.html etc files`

<1> What is `long_name` in this scenario ?
<2> What is the purpose of `encryption_key` field ?

The signature of this Structured Data can/will use some completely separate Key-Pair.
<3> Do we need to restrict the number of Key-Pair the use can generate or the number of Structured Data with type_tag = 4 that user can create ?

dirvine: So I think entries will be as you say, but found by looking for `SHA512(name + 3)` and ask fro type 3 data (i.e. the public dir). That will give you the `Directory` you want. 

1. Long name is really your name. So this `dns` packet would have `tag_type = 5` and `Identity = SHA512(spandan)` and if I look for your web site for instance I `Get(SHA512(spandan + 5)` and read the www entry (we demand at least www, blog, micro-blog etc. exist so these are required, developers can add more services, but we can set a minimum, tbd). So you will have one web site per identity by default, but you could have more if you add them as services and apps recognised those services. This is probably for later though. To access this all a client (person) will so is type `safe::spandan/www` or `safe::spandan/blog`into their web browser. Same for other services. This can be certainly improved (default to www for instance) 

2. This is currently un-used and can be removed for now. I will make that change. (it'e too much for now). 

3. We can restrict DNS entries if we wish, it is an implementation issue for sure. Good to get community feedback on this one. Personally I think we should restrict to 3 at a time (so require a delete of an old one to create a new one after that)

PR #21, opened 2015-07-02
Routing unified structured data

benjaminbollen:
Complements routings perspective on the RFC for Unified Structured Data (https://github.com/maidsafe/rfcs/pull/11).  This RFC is purely intended to elaborate on the practical impact for routing and allow the relelvant discussions limited to the implementation.  It is explicitly not intended to discuss the design; these discussions only belong on the parent RFC

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/21)


benjaminbollen:
should this be closed, merged to refused or accepted.  The work has been implemented, but not strictly as described here during Rust-3

dirvine:
I think this should be added as an addendum to the existing merged SD PR as it contains more detail and is helpful. If you can do that we can merge it in straight away as it does not argue or debate any features, just explains more.

PR #30, opened 2015-07-09
Unified structured data improvement

ustulation:


dirvine:
This is a very nice addition and simplifies an already simple structure (great) as well as ensuring collision proof stores. This is very good as a `Put` data will gi via another persona (ClientManager) and then onto store. Cross persona messages should not be allowed really so this is nice that collision errors are gone, means there is no known error for a `Put` of this data type.  :+1 from me (happily)

proposed/0006-unified-structured-data-imporvement.md
```diff
@@ -0,0 +1,147 @@
...
+```
+StructuredData {
+    other_fields: Value-22,
+    owner_keys: A-PublicKey,
+    signatures: Signed by A-PrivateKey,
+}
+```
+ 
+**step 7:**
+Vault will use owner field from **step 5** == `B-PublicKey` and signature verification fails. So it will not update anything.
```
inetic: I think in this case vault can reject the illegal update simply using the fact that `A-PublicKey` is not listed in `owner_keys` in the stored version.

dirvine:
Unless there are any opposition to this in next 12 hours I suggest we merge as active as this is a nice simplification and therefore increases security *by reducing confusion).

proposed/0006-unified-structured-data-imporvement.md
ustulation: This will not be full-proof or exhaustive though. Client-A could have sent:
```
StructuredData {
    other_fields: Value-22,
    owner_keys: B-PublicKey,
    signatures: Signed by A-PrivateKey,
}
```

proposed/0006-unified-structured-data-imporvement.md
```diff
@@ -0,0 +1,147 @@
...
+```
+ 
+**step 5:**
+Vault will check if new data is sent by valid owner. Use `owner_keys` from the stored data in step 3 to verify signature of data in step 4. If valid replace previous data and store new one:
+```
+StructuredData {
+    other_fields: Value-22,
+    owner_keys: B-PublicKey,
+    signatures: Signed by A-PrivateKey,
+}
```
dirvine: This may be where we get an issue. at the routing level checks will be made on this data for now a `Put` just needs to check at least one owner has signed the `Put` message (routing message).
  
For updates (`Post`) though in the routing layer then routing will check that the routing message is signed by one of the `PreviousOwners`  at the moment and pass the message up to vaults. If routing did not do this check then this scheme here would work, but routing may pass up invalid data? 
At the moment though it can be argued routing does only a basic check not a full check (it only checks a single signature) and the data may still fail in upper layers, however it does at least check this data is not being spammed by a non owner.

inetic:
I think there is one problem with this approach when there is multiple owners already of the StructuredData chunk. Say, the chunk currently stored in Vault looks like this:

```
StructuredData {
    other_fields: Value-1,
    owner_keys: [A-PubKey, B-PubKey],
    // Note, priv keys will not be transferred, this is just a notation to indicate
    // how the signature was created.
    signatures: [sign(A-PrivKey, Value11), sign(B-PrivKey, Value11)],
}
```
Now A-Client want's to change the data, so he sends:

```
StructuredData {
    other_fields: Value-2,
    owner_keys: [A-PubKey, B-PubKey],
    signatures: [sign(A-PrivKey, Value-2)],
}
```
Then Vault will replace A's signature with the old one:

```
StructuredData {
    other_fields: Value-1, // Still on value #1 because majority consensus did not take place yet
    owner_keys: [A-PubKey, B-PubKey],
    signatures: [sign(A-PrivKey, Value-2), sign(B-PrivKey, Value-1)],
}
```

Now notice that Vault has no proof that A-Client really owns the data with Value-1, because the signature `sign(A-PrivKey, Value-2)` signs the next data.

bcndanos:
I can see another problem with this scheme, where a StructuredData has two owners:
 
**step 0:**
Client-A sends:
```
StructuredData {
    other_fields: Value-00,
    owner_keys: A-PublicKey B-PublicKey,
    signatures: Signed by A-PrivateKey,
}
```

**step 1:**
Vault receives for the first time. So Just Stores it:
```
StructuredData {
    other_fields: Value-00,
    owner_keys: A-PublicKey B-PublicKey,
    signatures: Signed by A-PrivateKey,
}
```

**step 2:**
Client-B wants to add a new owner to this structured data:
```
StructuredData {
    other_fields: Value-00,
    owner_keys: A-PublicKey B-PublicKey C-PublicKey,
    signatures: Signed by B-PrivateKey,
}
```

But, at the same time, Client-A makes a change to the other_fields but he doesn't send the full list of owners(He doesn't know the change of Client-B):

```
StructuredData {
    other_fields: Value-11,
    owner_keys: A-PublicKey B-PublicKey,
    signatures: Signed by A-PrivateKey,
}
```

**step 3:**
Vault receives two requests for the same StructuredData. Which one wins the race?? 
```
StructuredData {
    other_fields: Value-¿¿00 or 11??,
    owner_keys: A-PublicKey B-PublicKey ¿¿C-PublicKey??,
    signatures: Signed by ¿¿A-PrivateKey or B-PrivateKey??,
}
```
I suppose that depends on the nonce sent by the clients, which means that one message will be discarded. But if the nonce sent is correct for both requests, it can be a lost of information.

---

Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/rfcs/30)

inetic:
I think this kind of situation where N __distinct__ owners can change the chunk in different ways is resolved by requiring a consensus of __majority__ of current owners, since there can't be two different majorities.

---







Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/rfcs/30)

proposed/0006-unified-structured-data-imporvement.md
inetic: > Then even though the PublicKeys in the owner fields would match, Signature still has to be checked and then rejected as illegal update. 

Yeah, I forgot to mention that apart from StructuredData, the real message also contains the PublicKey of the client together with signature of this message, routing will check this signature during parsing of the message so by the time the StructuredData reaches Vault/Client, you can be sure that it came from the given client.

> So signature check is the ultimate indicator of a legal/illegal update.

So, yes, but it happens implicitly in routing.

inetic:
Possible alternative (I know we discussed to leave this for later discussion, but I already had the nice ASCI art in place :-P, so didn't want to waste it :) ):
```
struct StructuredData {
    tag_type   : u64,
    identifier : crypto::hash::sha512::Digest,
    version    : u64,                         // mutable - incrementing (deterministic) version number
    data       : Vec<u8>,                     // mutable - in many cases this is encrypted

             +--> Map ensures only one (Signature, Option<Signature>) pair per one owner.
             |
             |     +--> Owner's identifier.
             |     |   
    owners : Map<PublicKey, (Signature, Option<Signature>)>
                              |                   |
                              |                   +--> Optional signature of the next version
                              |                        of the mutable fields
                              |
                              +--> Signature of the current version of
                                   the mutable data.    
}
```

ustulation:
Appears complicated: `owners` is a field that must be included while calculating the signature (else people can add fake owners). So for multi-sig, any addition to the Map by the next client will invalidate the signature of the previous client.

inetic:
> owners is a field that must be included while calculating the signature

Ah, yes, but it shouldn't be too much of a problem, when calculating the signature, you can just use the keys of the map. Pseudocode:

```
sign(PrivateKey, concatenate([serialise(version),
                              serialise(data),
                              serialise(owners.get_keys())]));
```

proposed/0006-unified-structured-data-imporvement.md
```diff
@@ -0,0 +1,177 @@
...
+```
+## Change in the Definition of GETs
+The above shows the cases for PUTs, POSTs and DELETEs. There will be a change in the signature of GETs to achieve the above no-collision scheme. GETs for StructuredData need to be changed from:
+```
+GET(Name, DataRequest::StructuredData(u64))
+```
+to
+```
+GET(Name, DataRequest::StructuredData(SHA512(PublicKeys)))
+```
```
maqi: u64 is better to be retained as it maybe used as a app_tag.
for example MpidMessaging may use structured_data for messages among nodes, and tag = 0x51000 indicates an MpidMessage meanwhile tag = 0x51001 indicates an MpidAlert

proposed/0006-unified-structured-data-imporvement.md
```diff
@@ -0,0 +1,177 @@
...
+For ordinary cases where `data-PUTers` themselves are fetching thier data (majority of clients) this does not mean much as the client engine will inject this silently as it already has the knowledge of user public-keys. The only time it does not have the knowledge is during the fetching of the session packet. The solution for this is easy as login packet has a dedicated maidasafe-type-tag and vaults can recognise it and make special allowance for it to be stored with `Key = Name` instead of `Key = (sha512(PublicKeys), Name)` as in above.
+
```
maqi: Please define such maidsafe-type-tag as a specific value. and is such maidsafe-type-tag is the type_tag field in structured_data?

proposed/0006-unified-structured-data-imporvement.md
```diff
@@ -0,0 +1,177 @@
...
+Since Dns Packets also have special type-tags, the vaults can recognise them. There could be ordinary ones (with TAG = X) where the storage is `Key = Name` for which there would be collisions and special ones (with TAG = Y) where the storage would be `Key = (sha512(PublicKey), Name)`. The safe-browser would land on the correct destination depending on whether TAG = X or TAG = Y was demanded. This is beyond the scope of this Rfc and can be discussed in a separate one.
+
```
maqi: I don't like the idea that the algorithm of key calculation needs to be based on type-tag.
We can make them consistent easily: if PublicKey is unknown, giving it a default value say [0u8; 64].

proposed/0006-unified-structured-data-imporvement.md
```diff
@@ -0,0 +1,177 @@
...
+# Alternatives
+None yet.
+
```
maqi: shall now rfc shall contain a section of required work for each lib?

maqi:
Is this is duplicated with the PR https://github.com/maidsafe/rfcs/pull/21/files ?

proposed/0006-unified-structured-data-imporvement.md
ustulation: Yes - it is the SD Tag field.

proposed/0006-unified-structured-data-imporvement.md
ustulation: It has to be based on type-tag at some level. Consider this: You are putting an SD onto the vaults - the vaults don't take tag into consideration. Every `PUT`ter has a valid MaidSafe Account and thus valid PublicKeys. So PublicKey will always be known for a `PUT` and vaults would never have to give a default value. However the public-key might  not be known for `GET`s for that particular SD - eg., fetching of a page by a browser - it wouldn't know the PublicKey of the user who `PUT` the SD Record. In that case if you default it to something, the algorithm's not going to work at all because you will never end up actually finding it.

proposed/0006-unified-structured-data-imporvement.md
ustulation: This was with Reference to the now outdated `DataRequest` as it now also contains `NameType` field. I'll update this section. But I still have some questions if you could answer. So present day `DataRequest`s are more like `DataRequest::StructuredData(id, tag)` .. when we route the `GET`s we already are calculating the `name` as `sha512(id + tag)` , so why would you additionally require tags `GET`s ? In `PUT/POST/DELETE`s you do have the entire SD with you as a payload and can thus extract the tag information if you need to and use that for some logic. So i don't get what you would miss `u64` was not retained.

maqi: there is use case that SD only being used as a wrapping tool for requesting.
For example, the MpidMessaging, the struct of mpid_message and mpid_head is not StructuredData, but the requesting of mpid_message and mpid_head is ustilizing current get/post/delete requests for SD, which in DataRequest(name, tag), tag will be used to differentiate the message or header.

ustulation:
@maqi 
> Is this is duplicated with the ...

No. That one was to help code the logic for the established RFC on SD. Referring to the Motivation Section there:
> The motivation for the current RFC is to establish a collective understanding of the changes needed for routing and sentinel to implementing the parent RFC.  This RFC explicitly excludes the intend to change
the design of the actual `StructuredData` and any such discussions should be posted on the parent RFC.

proposed/0006-unified-structured-data-imporvement.md
```diff
@@ -0,0 +1,177 @@
...
+StructuredData {
+    other_fields: Value-22,
+    owner_keys: B-PublicKey,
+    previous_owner_keys: A-PublicKey,
+    signatures: Signed by A-PrivateKey,
+}
+```
+ 
+**step 5:**
+Logic: Since `previous_owner_keys` in incoming data contains something, use `key` = `(sha512(previous_owner_keys), Name)` for look-up. Calculate a new `key` = `(sha512(owner_keys), Name)` if the update is legal as checked below (as usual).
```
dirvine: Where does the `Name` come from, perhaps needs clarified here?

PR #34, opened 2015-07-19
SafeCoin market price

bcndanos:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/34)

### DIFF
--- a/proposed/0000-safecoin-market-price.md
+++ b/proposed/0000-safecoin-market-price.md
@@ -1,7 +1,7 @@
-- Feature Name: (fill me in with a unique identity, my_awesome_feature)
-- Type (new feature, enhancement)
-- Related components (if any)
-- Start Date: (fill me in with today's date, DD-MM-YYYY)
+- Feature Name: SafeCoin_Market_Price
+- Type: New feature
+- Related components: maidsafe_vault, maidsafe_client
+- Start Date: 19-07-2015
 - RFC PR: (leave this empty)
 - Issue number: (leave this empty)
 
--- a/proposed/0000-safecoin-market-price.md
+++ b/proposed/0000-safecoin-market-price.md
@@ -9,9 +9,9 @@
 
 This RFC is based on the concerns I’ve read in the MaidSafe forum for the last six months. These are some of them:
 
-https://forum.safenetwork.io/t/lowest-put-price-discovery-and-new-safecoin-issuance/3819
-https://forum.safenetwork.io/t/cost-of-storing-data-on-safe/4285
-https://forum.safenetwork.io/t/the-price-of-safecoin-the-economics-behind-it/844
+* https://forum.safenetwork.io/t/lowest-put-price-discovery-and-new-safecoin-issuance/3819
+* https://forum.safenetwork.io/t/cost-of-storing-data-on-safe/4285
+* https://forum.safenetwork.io/t/the-price-of-safecoin-the-economics-behind-it/844
 
 This RFC is only another point of view (maybe not doable, feel free to reject this RFC) that breaks a lot the main design. Maybe this RFC can be decomposed in several RFC’s, smaller and feasible.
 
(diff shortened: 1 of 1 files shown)


proposed/0000-safecoin-market-price.md
```diff
@@ -0,0 +1,74 @@
...
+* https://forum.safenetwork.io/t/cost-of-storing-data-on-safe/4285
+* https://forum.safenetwork.io/t/the-price-of-safecoin-the-economics-behind-it/844
+
+This RFC is only another point of view (maybe not doable, feel free to reject this RFC) that breaks a lot the main design. Maybe this RFC can be decomposed in several RFC’s, smaller and feasible.
+
+
+# Motivation
+
+My first concern is how much cost to store a chunk of data (cost in terms of USD,EUR or BTC). No one is able to answer this question accurately. I’m thinking in a true Offer/Demand market, where you establish how much you want to pay to store data, and in the other side, the vaults establish how much they want to get paid to store that data.
+
```
dirvine: Ah this is where the network calculates in real time that cost. Otherwise it's not going to be able to give the most efficient price. It requires a lot of explanation, here is a massively simplified view of what happens and what's being tested. 

Data is stored in 3 locations (ignore replication etc.) Original Backup and Sacrificial
Sacrificial Data is the one that gets removed to store other types 

Sacrificial is also the network storing 50% more copies than it needs (very healthy) 

As each sacrifical is stored the Farming Rate doubles  (twice as hard to get a safecoin)
As each Sacrifical cannot be stored or is removed farming rate halves (farming more profitable)

Farming rate is then used to calculate cost of storage So a user will pay something close to the 5th root of farming rate per chunk (this is to be tested)

So this means the network calculates cost on network health and in `real time` which is very much designed to provide resources at the minimum price people (farmers) will provide those resources. Another push is farmer efficiency so people provide unused resources (computer already on share unused b/w/ hdd and cpu) which should drive costs of resource purchase to an absolute minimum. 

External safecoin pricing is separate but linked. If farmers get $X per hour they may wish to farm, if it goes to 1/X they may choose not to, the network will rebalance automatically regardless of external cost of safecoin. All it discovers is the farming rate required to get more resources, it does no know why that is the rate but calculates based on the rate required to attract/deter farmers whatever that may be.

proposed/0000-safecoin-market-price.md
```diff
@@ -0,0 +1,74 @@
...
+Another important aspect is the idea of pay once and store data forever. I can’t see how can be sustainable in the future. I’d like to see a system where you Put data into the Net as a renting service, like the Amazon Web Services.
+
```
dirvine: Yes this is another area questioned a lot. The notion is farmers must farm *above* the existing amount of data to get rewarded. 

The network has no idea of time and cannot recognise people, certainly not via a payment.  It could with safecoin (perhaps) but then it would need to link a persons data in an account (security issues) which the network holds and can recognise their data.

The network also does not know the notion of time in any sense, it does not use time anywhere except for very local calculations (never transmitted). Time is alien to the network protocol. 

For safety therefor all data is stored forever, unless signed by people.

Signed data is another mechanism being introduced at the moment (RFC in progress) where immutable data can be deleted. This will be more expensive that normal (de-duplicatable data) but can be deleted (by owner).  The owner is basically the private key holder of an embedded public key and will not leak any info about actual identities. 

Another area not covered yet is archive nodes, these will take older data and make them less stressful on the network by creating superchunks, these superchunks will contain millions of chunks in a structure that allows easy (but slower) retrieval of such data. The archive nodes will be paid at a higher rate we imagine that normal farmer nodes. RFC to come though. Dev bundle 3  first :-)

bcndanos:
OK. It seems you have all thought. Thank you for the answers. 
I close this PR.

proposed/0000-safecoin-market-price.md
a-jamesregan: @dirvine, does this mean that when everyone deletes a chunk of data stored on a network that it does not delete?  For example, if I store a piece of data that only I would have (such as a personal photo) and then I delete it, it won't automatically be removed from the network when no one has the data anymore?

dirvine: Yes that's correct, otherwise the network would have to recognise all data (which is a major security issue) There is an RFC  coming for deletable data, this will be much more expensive than normal de-duplicate data though.

PR #36, opened 2015-08-05
Increase limits on file sizes in self_encryption.

brian-js:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/36)


proposed/0000-template.md
```diff
@@ -0,0 +1,30 @@
...
+
+# Motivation
+
+A significant portion of network traffic is anticipated to result from the storage/retrieval of data coming through self encryption. It is therefore necessary to have robust, secure and efficient methods in place for handling the data arriving at any connected device. Tied to this are the parameters associated with the devices themselves, for our purpose mainly storage and memory capacities. By introducing a memory map we can increase the size of files that can be handled by self_encryption.
+
+# Detailed design
+
+Currently all data associated with a DataMap going through self encryption is held unencrypted in memory. A close call on a SelfEncryption, SE, object initiates encryption and storage of the data via the Storage object passed on creation. In the current iteration it is proposed for all systems to maintain a vector of data per file up to a 50Mb limit in memory, while for files that grow larger than this to introduce an anonymous memory map up to 1Gb on 32-bit systems and 10Gb on 64-bit systems.
+
```
dirvine: Could we explicitly detail all architectures here and the limit we will place on them? i.e. arm 7 10Gb possibly allowing several hundreds of files up to that size (depending on machine resources). ARM 64 total 256Tb etc. 

Also worth noting up to 50Mb will be handled with RAM and passed off after this, so we anticipate any file number limits to be well within any acceptable bounds.

proposed/0000-template.md
```diff
@@ -0,0 +1,30 @@
...
+# Drawbacks
+
+The 1Gb limit on file size for 32-bit systems is restrictive, however, this will be addressed in some future RFC.
+
+# Alternatives
+
+Allow unrestricted file sizes across devices by introducing a cache backed sliding window algorithm for self encryption.
```
dirvine: Maybe describe this a little, so a 500Mb sliding window perhaps (maybe less on 32 bit x86)

dirvine:
I think the rfc for increase self_encryption file size is good, but linked to another rfc in this PR. So I suggest close this PR and replace with 2 separate PR's @brian-js

PR #37, opened 2015-08-07
SAFE browser plugin URL handling / dynamic HTML on client side

theWebalyst:
I've been thinking about the plugin support needed for dynamic web apps and the result is this suggestion for how the browser plugin could handle URLs and enable client side scripting of dynamic HTML pages.

Death to the server!

### DIFF
--- a/proposed/0000-browser-plugin-URLs.md
+++ b/proposed/0000-browser-plugin-URLs.md
@@ -1,0 +1,95 @@
+- Feature Name: Browser Plugin URL Handling
+- Type: New
+- Related: SAFE Browser Plugin
+- Start Date: 07-08-2015
+- RFC PR: (leave this empty)
+- Issue number: (leave this empty)
+
+# Summary
+
+A method of URL handling and an extension to the SAFE Browser Plugin API, that will allow dynamic HTML pages to be supported in an efficient manner, entirely on the client side. It includes a method to GET the client side HTML/Javascript document concurrently with GET of other documents needed to render the page, to speed up page loading. Also for within-page bookmarking using fragment identifier.
+
+# Motivation
+
+This functionality is required in the plugin so that the client side can do things that are normally handled on the server when generating HTML pages dynamically. 
+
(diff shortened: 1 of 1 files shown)


dirvine:
@theWebalyst IS this still and active proposal Mark? If so we should try and progress a discussion in the next couple of weeks.

theWebalyst:
I'm not sure. I think it depends :-)

From memory I was trying to ensure that a URL could work along the lines of a dynamic website, for example:
```
http://domain.tld/some-path-to-a-template?post=123
```
I'm not sure if this is what is proposed, but think the existing launcher API might use # instead of ? as in:
```
http://domain.tld/some-path-to-a-template#post=123
```

In which case this causes a SAFE REST request for ```http://domain.tld/http://domain.tld/some-path-to-a-template``` (ie the URL up to but excluding the # and what follows), while the client side (JavaScript) can read the whole URL and so make a separate request for the appropriate content (by constructing a separate SAFE REST based on "post=123").

If I've understood that correctly, this is functionally fine and quite a neat solution. But one of the issues I was concerned about was page load times, particularly on first load. It remains to be seen what the latency is on SAFE network, and therefore how much of an issue this is, but the scheme as I understand it would mean a minimum of two sequential requests which effectively doubles the latency for a page that isn't in the local browser cache. I'm concerned that this will compare poorly with a current dynamic website where the whole URL generates a single request that is responded to by the server. 

Obviously there are also many websites which load content dynamically (AJAX style) but I don't know enough about all this to know how these might compare.

Anyway, my scheme called for the ability to initiate multiple concurrent SAFE REST requests within a single URL, with the response data available in the browser client API (e.g. using promises). Doing this is not very elegant, so if SAFE network latency isn't an issue the '#' scheme would be preferable (simpler to understand and implement). But if latency means that dynamic websites are noticably slow to load I'm concerned about the impact this would have on their usefulness, SAFE adoption and the perception of the project. So, it depends :-)

dirvine:
I see what you mean here Mark. Lets leave this in place for a while longer and wait till we get some real world tests of response times when everyone has vaults in place. We can revisit this then and confirm there is an issue and if so use this as a place to bounce further ideas off of. Is that OK with you?

theWebalyst:
Yep, makes sense. Can't wait to see the thing fly :-)

PR #38, opened 2015-09-11
Launcher RFC - initial commit

ustulation:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/38)


proposed/0006-Launcher.md
```diff
@@ -0,0 +1,192 @@
...
+
+**step 2:** If it was a log in, Launcher Fetches and decodes User-Session-Packet (USP).
+
+**step 3:** Launcher Fetches Maidsafe Specific Configuration Private Root Directory ID (See Session Packet description) - if not present Launcher will Create it (via [safe_nfs](https://github.com/maidsafe/safe_nfs) crate)
+
+**step 4:** Launcher Reads the special Directory reserved for it. (See Session Packet Description) - if not present Launcher will Create it (via [safe_nfs](https://github.com/maidsafe/safe_nfs) crate)
+
+**step 5:** Launcher Fetches User’s Private root Dir - if not present Launcher will Create it (via [safe_nfs](https://github.com/maidsafe/safe_nfs) crate)
+
+**step 6:** Launcher Checks for special Directory named `SAFEDrive` inside the user’s root directory - if not present Launcher will Create it (via [safe_nfs](https://github.com/maidsafe/safe_nfs) crate)
```
Fraser999: I think this would be better as `SAFE Drive` (or if we don't want to handle spaces, then `safe_drive`).  Same for other file/folder names mentioned below - e.g. `LauncherReservedDirectory` -> `Launcher Reserved Directory` and `DnsReservedDirectory` -> `DNS Reserved Directory`.

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Start Flow
+
+**step 0:** Start Launcher.
+
+**step 1:** Enter Credentials to either create an account or to log into a previously created one.
+
+**step 2:** If it was a log in, Launcher Fetches and decodes User-Session-Packet (USP).
+
+**step 3:** Launcher Fetches Maidsafe Specific Configuration Private Root Directory ID (See Session Packet description) - if not present Launcher will Create it (via [safe_nfs](https://github.com/maidsafe/safe_nfs) crate)
```
Fraser999: "Maidsafe" -> "MaidSafe"

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,192 @@
...
+```
+- Create/Append to local config file (on the user's machine) the following:
+```
+[
+    { App-ID, “/path/to/XYZ” }, // This is newly added by the above
+    { App-ID, “/path/to/some/other/app/already/there” },
+    ... etc.,
+]
+```
+- The format of the config file will be CBOR (compact-binary-object-representation). The name of the local config file should be `<LOCAL-CONFIG-FILE> = .launcher-local-config-file`. The config file location flowchart shall be same as that of `crust` crate's.
```
Fraser999: The name of the Crust config file is `<app_name>.crust.config` - I think it would be good to stay consistent with that.

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,192 @@
...
+```
+struct Request {
+    prefix: u8,  // This must be ASCII of '?' (a question mark)
+    nonce : u32, // This must be the nonce supplied by Launcher
+}
+```
+
+**step 6:** Launcher reads the `<LAUNCHER-CONFIG-FILE>` and creates a mapping in the MaidManagers for associating user’s `MAID Keys <-> App specific Keys` to allow App to PUT/POST on behalf of the user in `<APP-ROOT-DIR>`.
+- The mapping is done only for the ownership key/s (not encryption keys).
+- The request for mapping/un-mapping shall be a command to the `MaidManagers`. For commands to `MaidManagers`, `StructuredData` with special Type-Tag will be reserved. On reception of this `StructuredData` the vaults will check the payload and act upon the request instead of executing a normal reaction to the usual `PUT/DELETES`.
```
Fraser999: From the Vault perspective, I think this would be better suited to POST commands rather than having to handle a special case for PUT and another for DELETE.  It should keep the Vault code logically cleaner.

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,192 @@
...
+- `StructuredData` with Tag-Type **9** will be reserved for `Client <-> Vault` messages.
+- The messages will be defined in [safe_vault](https://github.com/maidsafe/safe_vault).
+- This shall be the format of `Client <-> Vault` messages:
+```
+pub enum ClientVaultMessage {
+    MapKeys {
+        ownership_key_to_map   : sodiumoxide::crypto::sign::PublicKey, // Launcher-Assigned-Public-Key-For-App
+        ownership_key_to_map_to: sodiumoxide::crypto::sign::PublicKey, // User-PublicMaid-sign-key
```
Fraser999: This is the key the Launcher's already using isn't it?  Do we need to duplicate that here?  Same questions for `UnMapKeys` below.

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,192 @@
...
+struct Response {
+    root_directory_id    : NameType,
+    root_directory_tag   : u64
+    public_signing_key   : sodiumoxide::crypto::sign::PublicKey,
+    private_signing_key  : sodiumoxide::crypto::sign::SecretKey,
+    public_encrytion_key : sodiumoxide::crypto::box_::PublicKey,
+    private_encrytion_key: sodiumoxide::crypto::box_::SecretKey,
+}
+```
+
```
Fraser999: We had planned this before to have the app generate its own keys and pass them to the Launcher.  This differed in two major ways.
1. The Launcher never gets to see the app's private keys and they're never transmitted between these processes.  They're not stored in any config file anywhere - neither by the Launcher, nor the app.
2. The keys then are session keys.  They only need to be passed to the Vault network by the Launcher - not deleted.  The app itself can connect to the network at the same location as the Launcher, and if the Launcher has provided the app keys, the app can PUT/POST/DELETE.  Once the app disconnects from the network, the Vaults holding the app keys will know about this without needing to be told, and will remove them.

Having these as session keys is useful not only because it limits the time for which they are usable, but also it acts as a disincentive for the app to try and store them somewhere (e.g. unencrypted on the native filesystem).  It also dissociates the App and Launcher even further, since the Launcher doesn't have to be running in order for these app keys to be removed from the network.

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,192 @@
...
+## Misc
+- If the App is added to Launcher in one machine, the mention of this will go into `<LAUNCHER-CONFIG-FILE>` as stated previously. It will thus be listed on every machine when user logs into his account via Launcher on that machine. However when the App is attempted to be activate on a machine via Launcher where it was not previously added to Launcher then he will be prompted to associate a binary. Once done, the information as usual will go into the `<LOCAL-CONFIG-FILE>` on that machine and the user won't be prompted the next time.
```
Fraser999: "activate" -> "activated"

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,192 @@
...
+- When the App is started via Launcher, it will first check if the `SHA512(App-Binary)` matches any one of the corresponding entries in the vector in `<LAUNCHER-CONFIG-FILE>`. If it does not Launcher will interpret this as a malacious App that has replaced the App-binary on user's machine, and thus will not pass any credentials to it nor communicate with it.
```
Fraser999: It's more likely that the app has been updated here isn't it?  I guess we should just seek user-confirmation of the change while warning them about the chance of malice?

proposed/0006-Launcher.md
ustulation: `POST`s don't go to Maid-Managers I think

proposed/0006-Launcher.md
ustulation: I would be against spaces in names created by the library (ie., without user intervention). Spaces is more sort of non-default in systems like Linux where you need to quote or escape them, both of which are unnatural (eg., `cd Dir\ With\ Spaces` etc)

proposed/0006-Launcher.md
ustulation: So should it be modified to `launcher.config` ?

proposed/0006-Launcher.md
ustulation: No, each app gets its own unique key that'll be mapped to the user's Maid keys. Once the app starts it does not require launcher to be on/running as a background process. When the App does a PUT with it's owner keys the MaidManagers should be able to see that as done on behalf of the user and subtract his balance etc. So the necessity of mapping.

proposed/0006-Launcher.md
ustulation: Do you mean the App deterministically generates these keys ?

if yes then: I think randomly generated Maid keys are better and hence we use that stratergy in Login Packets - the Login Packet is encrypted with deterministic keys but the Maid Keys it contains which eventually is used to mutate the network is non-deterministic for every client and hence stored in Login Packet.

if not then: these will need to be stored somewhere

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Misc
+- If the App is added to Launcher in one machine, the mention of this will go into `<LAUNCHER-CONFIG-FILE>` as stated previously. It will thus be listed on every machine when user logs into his/her account via Launcher on that machine. However when the App is attempted to be activated on a machine via Launcher where it was not previously added to Launcher then he/she will be prompted to associate a binary. Once done, the information as usual will go into the `<LOCAL-CONFIG-FILE>` on that machine and the user won't be prompted the next time.
+- When the App is started via Launcher, it will first check if the `SHA512(App-Binary)` matches any one of the corresponding entries in the vector in `<LAUNCHER-CONFIG-FILE>`. If it does not Launcher will interpret this as a malacious App that has replaced the App-binary on user's machine, and thus will not pass any credentials to it nor communicate with it.
```
Fraser999: It's more likely that the app has been updated here isn't it? I guess we should just seek user-confirmation of the change while warning them about the chance of malice?

proposed/0006-Launcher.md
Fraser999: Yeah - I know it's non-default on Linux, but on Windows and OS X it's very normal.  I'd rather go with what the majority of OSs use than the minority.

proposed/0006-Launcher.md
Fraser999: I think so.

proposed/0006-Launcher.md
Fraser999: You're correct, but Vaults also don't currently handle any special cases of PUT/DELETE either, so there's changes to be made either way.

However, I guess your point is more that my suggestion would impact Routing as well as Vaults?  I do feel that using a PUT here is slightly in conflict with how I saw the network working (I'd anticipated "special actions" to be sent via POST messages), but I'm probably viewing it wrong again, and we can definitely make it work at Vaults as you suggested.

proposed/0006-Launcher.md
Fraser999: Yeah - it was the Maid key I was talking about, not the app key.  The Launcher's already connected to the network and using that Maid key isn't it?  So the MaidManagers know which Maid the launcher's trying to map - they only need told about the app key.

proposed/0006-Launcher.md
Fraser999: No - I meant the app generates non-deterministic, asymmetric keys per session.

It then sends the public key to the Launcher via the IPC channel, and the Launcher passes that key to its MaidManagers.  The launcher doesn't keep hold of that key.

The MaidManagers expect the app to connect to the network there, and the app can prove it's the owner by signing a non-deterministic query from them.  They then hold on to the app's public key (and allow it to act on behalf of the corresponding Maid) for as long as the app stays connected.

The app has no incentive to try and store the keys anywhere itself, since the keys are only useful while it's running during that session.

proposed/0006-Launcher.md
ustulation: Ah i see, so you want the 2nd variable gone. Yes that appears a valid point. I'll let it linger around for sometime though before updating the PR so others could comment if they have something to add.

proposed/0006-Launcher.md
ustulation: Normal does not correspond to best-practices always. Consensus (though not a deep search by me) seems to no use spaces (to which i would strongly agree especially if it's a program generated file/directory):
https://library.stanford.edu/research/data-management-services/data-best-practices/best-practices-file-naming
http://www.exadox.com/en/articles/file-naming-convention-ten-rules-best-practice

And even for OSX by the OSX people:
https://support.apple.com/en-us/HT202808

proposed/0006-Launcher.md
ustulation: That wouldn't work unless more stuff is modified. The App can `PUT` StructuredData in one session but can neither `POST` or `DELETE` it in subsequent sessions as you will have a new set of ownership keys for every session (and in every machine) and lost the ownership of the previous data.

proposed/0006-Launcher.md
Fraser999: My point is mainly about ease-of-use for end users and aesthetics.  There are many well-known default files and folders on all versions of Windows and OS X which contain spaces.  So even if there _are_ best-practice documents, they aren't adhered to by the OS providers themselves.

proposed/0006-Launcher.md
Fraser999: OK.  The last time I was involved here, the Maid was to be the owner of the data, all client requests went through the MaidManagers, and we weren't planning on signing requests.  Hence my confusion.

I still like the idea of the session keys for the two main reasons I listed, and I guess we could make that work, but that'll inevitably make the process (and hence code) more complex, so it's probably a non-starter.

ustulation:
Raising a revamped one - closing this

PR #39, opened 2015-09-14
As Directories are identified by a tuple of 4 parameters (NameType, t…

ustulation:
…ag, if private/encrypted, if versioned) it makes sense to have the 4 identifies stored for better scalability.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/39)


dirvine:
Cool Spandan, we can RFC improvements here for sure. Thx

PR #40, opened 2015-09-21
Address relocation proposal.

brian-js:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/40)


proposed/0007-Address_relocation.md
```diff
@@ -0,0 +1,40 @@
...
+
+In it's current form the process ensures that a joining node cannot occupy a position of it's own choosing. However, in a semi-stable network state it may be possible given the joining node knows the two closest nodes, N3, N4 say, in it’s joined group to generate keys that are relocated close to some identifier, I say, and thus obtain group influence at that location. It is the purpose of this RFC to avoid that possibility by storing the two close node identities, N3 and N4, at each node belonging to the joined group in order to reject relocations involving either of those nodes if they occur more than once for the participating nodes in a given session.
+
+## Implementation
+
+The `RoutingNode` function `handle_request_network_name` calculates a relocated name for a node joining the network. The utils.rs function `pub fn calculate_relocated_name(mut close_nodes: Vec<::NameType>, original_name: &::NameType) -> Result<::NameType, ::error::RoutingError>`, takes the close nodes from the routing table and original name, K, as arguments, returning H(K + N1 + N2) = N, in the notation above, and is used for this purpose.
```
maqi: The Mpid Messaging proposal (https://github.com/maidsafe/rfcs/pull/43) requires Address Relocation having a capability to grant a determined connecting address based on account to client.
Will this be considered?

brian-js: It's not considered with this rfc it concerns non-client nodes only.

proposed/0007-Address_relocation.md
```diff
@@ -0,0 +1,40 @@
+- Feature Name: Prevent any single address being used more than once during address relocation.
+- Type enhancement
+- Related components routing, safe_vault
```
maqi: according to the description below, it seems the implementation can be achieved solely in routing lib.
As I understand, this address relocation shall also be done solely by routing lib (unless it also covers client node, which client lib may need to be involved to provide account info to calculate connection address).
So, I think vault shall be removed as a related component.

brian-js: Yes. It affects vaults but doesn't directly affect the code. I'll change it.

proposed/0007-Address_relocation.md
maqi: A pmid_node node is a non-client vault playing the role of storing immutable_data.
The current implementation assumes the network address of a pmid_node is it's associated account and will be stored on its correspondent PmidManger group.
This means the network address of pmid_node needs to be determined by account, which is conflict with the dynamic requirement for vault's other roles.

One possible solution is to make pmid_node a separate vault instance using pre-determined network address, meanwhile other personas kept in another vault instance using dynamic-determined network address.

However, this seems more suitable to be covered by Issue https://maidsafe.atlassian.net/browse/MAID-118 instead of in this rfc

Fraser999: > One possible solution is to make pmid_node a separate vault instance using pre-determined network address derived from account, meanwhile other personas kept in another vault instance using dynamic-determined network address.

I think this is the best approach.  That Vault doesn't even need to be the same process, or even the same GitHub project.  In some regards the PMID persona is more like a client anyway: it's a single entity on the network which can only validate its actions by signing - it's not a member of a group of nodes.  So if it has no authority to mutate data, there's no need to ever apply address relocation to it.

In fact, running it as a separate process would probably be useful for low-power, low-spec, or bandwidth-limited scenarios, since such a device could possibly run the personas Vault quite comfortably (and hence be "giving back" to the network) and not run the storage Vault.

Another possible solution is that relocation is permanent - i.e. the network-generated value `N` forms part of the PMID packet for the new node.

This means the relocation only happens once when a new Vault joins for the first time.

The downside with this is the longer the Vault exists, the less "validatable" `N` will become - the nodes which created `N` will become lower and lower in the list of closest to the original address as the network grows.

maqi: The downside of having pmid_node using in a separate process is :  such node's shall appear as `client` to routing, but DataManager person still need to have the knowledge of close clients as pmid_node candidates.
This enforce churn_client and churn_vault to be both reported to Vault level.

Fraser999: Yeah - this has been a problem for a long time: PMID node is neither fully a Vault nor a Client - it is a bit of both things.

maqi: I will then prefer making pmid_node a separate lib using routing_client with a specific type_tag (routing shall have an API change to includes a tag field allowing different kinds of client to connect and being handled differently). 

rules for routing will still be simple : 
1, such routing_client is still a client, i.e. shall not appear in routing_table and being involved in  any close group calculation for vault
2, In case of churn happens to a peer and it is a routing_client with special_tag, such peer shall be reported to vault

proposed/0007-Address_relocation.md
```diff
@@ -0,0 +1,40 @@
+- Feature Name: Prevent any single address being used more than once during address relocation.
```
maqi: The file name doesn't need to contain a prospective RFC number.
Please DO NOT commit a change to fix it before all the other comments got resolved, otherwise the comment history will be lost

proposed/0007-Address_relocation.md
```diff
@@ -0,0 +1,40 @@
...
+- Type enhancement
+- Related components routing
+- Start Date: 21-09-2015
+- RFC PR: (leave this empty)
+- Issue number: (leave this empty)
+
+# Summary
+
+Prevent a potential attack where a node gains overall/majority influence in a group setting.
```
maqi: `node` here may give a wrong impression that the network allows a node having more weight.
I think it is more accurate to say: ... attack where a hacker spawn clustered nodes closing to each other to gain overall/majority ...

proposed/0007-Address_relocation.md
```diff
@@ -0,0 +1,40 @@
...
+
+# Motivation
+
+The possibility 'may' exist for a node on the network to generate off-line derived identities close to a specific target and to then connect nodes whose resultant network relocated identities actually lie close to that target. In such cases the owner of that node will dominate, or at least have a dominant influence within, the group close to that target. This RFC proposes an amendment to the current process to prevent this possibility arising in practice.
```
maqi: `node` here shall be `attacker` ?

proposed/0007-Address_relocation.md
```diff
@@ -0,0 +1,40 @@
...
+
+Having computed N, each node in the group H(K) sends a message, containing K, to the close group of N to confirm address relocation has been applied for the node owning K. A sufficient condition for the receiving group to resolve the message is for a quorum in the sending group to agree on the two closest nodes. Each node in the group H(K) further sends a message, containing N, to the joining node confirming the network address it now occupies.
+
+In it's current form the process ensures that a joining node cannot occupy a position of it's own choosing. However, in a semi-stable network state it may be possible given the joining node knows the two closest nodes, N3, N4 say, in it’s joined group to generate keys that are relocated close to some identifier, I say, and thus obtain group influence at that location. It is the purpose of this RFC to avoid that possibility by storing the two close node identities, N3 and N4, at each node belonging to the joined group in order to reject relocations involving either of those nodes if they occur more than once for the participating nodes in a given session.
```
maqi: what does `semi-stable network state` mean?
Say an attacker having the first faulty node joined (via closing peer N1&N2 and K1) and  knows a closing peer N3&N4, then he can starts calculating a key so that `H(K + N3 + N4)` will be close to the target.
I understand the paragraph here is trying to avoid such attack scenario, but the chance that N3&N4 had been used previously (so get stored in the targeting group ) is no higher than the normal joining procedure. So I don't see the point how it can prevent such attack and how a normal joining procedure won't get blocked by this proposal.

proposed/0007-Address_relocation.md
```diff
@@ -0,0 +1,40 @@
...
+- Issue number: (leave this empty)
+
+# Summary
+
+Prevent a potential attack where a node gains overall/majority influence in a group setting.
+
+# Motivation
+
+The possibility 'may' exist for a node on the network to generate off-line derived identities close to a specific target and to then connect nodes whose resultant network relocated identities actually lie close to that target. In such cases the owner of that node will dominate, or at least have a dominant influence within, the group close to that target. This RFC proposes an amendment to the current process to prevent this possibility arising in practice.
+
```
maqi: I think the main purpose of using address-relocation is to avoid `user defined identity` to be used as vault's address.
The complexity of attack under `K closer to T` and `Hash(K + N1 + N2) closer to T` is the same and will be easy to achieve, as long as user provided identify can be used (i.e. attack modify the routing's ID generating algorithm to using fixed K).
Such attack can only be avoided only when a joining node's address is granted by the network dynamically.

proposed/0007-Address_relocation.md
```diff
@@ -0,0 +1,40 @@
...
+# Detailed design
+
+A node joining the SAFE network generates a new cryptographic key pair to access the network for each session. The generated public key, K, is then cryptographically hashed using some hash function H to give H(K). The node then sends a message, containing K, to the group H(K) requesting a network identifier for the session. On receipt of such a request each node belonging to the group hashes the concatenation of K with the two closest nodes, N1, N2 say, in xor distance from K present in their routing tables, giving H(K + N1 + N2) = N, the identifier the node will use on the network for the session.
```
maqi: The complexity of attack under `K closer to T` and `Hash(K + N1 + N2) closer to T` is the same

proposed/0007-Address_relocation.md
brian-js: It means long enough for a would be attacker to run the calculation before the routing tables of the target group are significantly changed. H(K) would be close the joined group. We're avoiding the case if H(K + N + M) for N, M in the close group relocates close to I.

proposed/0007-Address_relocation.md
brian-js: The node generating identities offline will be a connected node.

proposed/0007-Address_relocation.md
benjaminbollen: A PmidNode is only a PmidNode after it has connected into the network.  Your question was sufficiently answered by @brian-js' initial response: this concerns non-clients, also known as full nodes.  Full nodes are not persistent, so a PmidNode is only a ManagedNode authority.  You should make your claim that "vault requires this PmidNode to be allowed to connect to a desired location" more clear, or this remark can be fully considered closed.

benjaminbollen:
The motivation for this RFC is to improve the ability for the network to fend off an offline attack on a targeted group.  Address relocation is already implemented and the issue at hand is that it is not activated - as activating the feature would block existing nodes to connect in the current form of connection management.  This leads to the conclusion that the RFC to "activate the enforcement of Address Relocation" is best done after the RFC for connection management review.

A very brief overview on Address Relocation as it is implemented, can be compared by the following back of the envelope reasoning:

1. to establish a node close to a target, without Address Relocation, one has a chance of 2^(-n) to find a close node from a generated public-private key pair; where n is the number of leading bits that need to be identical to the target in order to be considered close.

2. with address relocation, as implemented, it uses network state (the closest nodes to the H(K)) to uniformly redirect the connect location.  The problem that aims to be solved is that a node might be able to observe what this network state is.  This is indeed possible, but it does not change the probability improvement that address relocation in the current form already offers: an attacker needs to find a key-pair K such that H(K) is close to the known network nodes, AND that the H(H(K) + n1 + n2) is close to the target.

This means that for a small effort on the networks side, the attacker is now forced to search for a key with a probability of 2^(-2n), exponentially worse!  The number of n is determined by the network size, and roughly scales with the logarithm of the number of nodes - the crux of the SAFE network: the large numbers provide local security.

In a lesser matter, the churn on the network limits the timeframe in which an attacker can search for this key pair.

I will address problems with the RFC proposal in a next comment

benjaminbollen:
Problems with the current RFC:
1. it is proposed that blocking re-use of the nodes used to relocate a node can abide the problem:
1.a. An attacker can repeat this offline key-searching on different groups independently, fully bypassing the blocking effort
1.b Blocking the use of two node names to relocate a single new node has a declining population: if it takes two nodes to generate a single new node, then your population size will decrease to zero; or never reach two if starting from one.

2. retaining state of the used node_names (to be used to block re-use) does not address the actual problem: namely that the information used to network-relocate a new node needs to contain an on-request-generated-(pseudorandom) input.  The closest nodes were take in an initial design as an approximation (given churn) pseudo-random input.  The clear and only improvement is improving the pseudo-random nature of that input information.  It needs to be repeated very clearly that even without such improvement the value of Address Relocation in the current form is a sensible protection feature to have as a fixed small effort on the network induces an exponential suppression of the attack vector.

benjaminbollen:
As a summary:
1. the existing Address Relocation mechanism is a sensible and already implemented defense measure.  It is however not ACTIVATED
2. Activation of Address Relocation requires that existing nodes can get a group confirmation of their legitimate participation in the network - this was the RFC I understood under Address Relocation
3. Address Relocation at a later stage can be improved, and truly make offline generation attacks impossible by generating pseudo-random input into the calculation at the moment of the request.  The current input of the close-nodes is an imperfect approximation, but it does not erode the value of the mechanism

benjaminbollen:
@maqi @Fraser999 regarding the concerns raised for PMIDnodes, address relocation takes place before the vault personas are initiated, so the pmidnode name is stable during its lifetime.  Vaults are currently non-persistent and so should be the chunk store for PMIDnodes.  Future plans for making "archive nodes" persistent would validly raise the question "how a node that comes back online can from a newly appointed location, proof valid ownership of the data he stored while offline".  This is currently outside the scope of our features

maqi:
@benjaminbollen : I know address relocation happens before personas and keeps stable during the session. However, a `user account` persists real lifetime and need to link his pmid_node to the account to gain the credits for providing pmid_node service.
With a changing address, the above can only be achieved by `register` pmid_node with an account, however to secure and verify the ownership will cause design/implement issues.
Also, a changing address to pmid_node means the chance of `refreshing` an im_data will increase a lot, which will be a burden to the network.

canndrew:
> This means that for a small effort on the networks side, the attacker is now forced to search for a key with a probability of 2^(-2n), exponentially worse!

This is only quadratically worse. With C nodes on the network the difficulty goes from O(C) to O(C^2) (or, O(2^(-n)) to O((2^(-n))^2)).

canndrew:
In practice though, the difference between n and n^2 can be the difference between feasible and infeasible.

benjaminbollen:
cheers, you are right ;)

benjaminbollen:
> @benjaminbollen : I know address relocation happens before personas and keeps stable during the session. However, a user account persists real lifetime and need to link his pmid_node to the account to gain the credits for providing pmid_node service.
With a changing address, the above can only be achieved by register pmid_node with an account, however to secure and verify the ownership will cause design/implement issues.
Also, a changing address to pmid_node means the chance of refreshing an im_data will increase a lot, which will be a burden to the network.

Over the lifetime of the pmid_node, the node earns safecoin.  That earned balance is persistent and independent of the pmid_node.  When the computer goes offline, it will have to restart with a new PmidNode name and zero history.  It can earn into the same wallet though.

PR #42, opened 2015-09-24
Launcher as a service

ustulation:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/42)


proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
+- Feature Name: SAFE-Launcher
```
Fraser999: To disambiguate it from the "Launcher-Simplified" RFC, this should be renamed to something like "Launcher as a service/daemon".

Fraser999:
The file name shouldn't contain the prospective RFC number.  Also, it should be renamed to match whatever the updated "Feature Name" becomes.

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+- Type: New Product
+- Related components: [safe_client](https://github.com/maidsafe/safe_client), [safe_nfs](https://github.com/maidsafe/safe_nfs)
+- Start Date: 11-September-2015
+
+# Summary
+
+Launcher will be a gateway for any App that wants to work on the SAFE-Network on a user's behalf.
```
Fraser999: This needs fleshed out a bit.  It should still be a single paragraph, but it should allow third-party readers to get a rough idea of what the product will be, and this also should allow readers to disambiguate between the other proposed Launcher RFC.

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+
+# Motivation
+
+## Why?
+
+App's access of the SAFE-Network on behalf of the user is an issue with high security concerns. Further, decentralisation allows for uniformity in experience on various machines concerning user installed Apps.
```
Fraser999: First word: "App's" -> "Apps"
"SAFE-Network" -> "SAFE Network" (multiple instances in file)
last word: "Apps" -> "apps"

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+
+## What cases does it support?
+
+SAFE-Launcher
+
+<1> will allow user to create an account and/or log into the SAFE-Network.
```
Fraser999: Use markdown for all lists: either:
```
1. one
1. two
```
yielding:
> 1. one
1. two

or
```
- one
- two
```
yielding
> - one
- two

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+
+<2> will authenticate a user installed App to access SAFE-Network on the user's behalf.
+
+<3> will manage metadata related to Apps to give uniformity in experience when shifting from one machine to another - eg., if App `A` is installed in machine 1 then when the user logs into machine 2 using his/her SAFE-Account via Launcher, he/she will be presented with a union of all the Apps that were installed on all the machines which access the SAFE-Network on his/her behalf.
```
Fraser999: "eg.," -> "e.g."

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+
+<4> Not allow Apps to mutate each other's configs.
+
+<5> Easily Revoke App's ability to Read and Mutate the Network on User's behalf.
```
Fraser999: most capitalised words here shouldn't be.

proposed/0006-Launcher.md
Fraser999: Also, bullet list items generally don't require a capitalised first word, nor a full stop.  Either way, we should be consistent with both aspects.

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+
+## Expected outcome
```
Fraser999: This whole section is a duplicate of the one above.  I'd suggest either making it meaningfully different, or removing it.

proposed/0006-Launcher.md
Fraser999: I think it would be good to expand a bit on the security concerns (a couple of examples maybe).

Also, isn't another motivating factor UX, in that apps won't need to get the user's credentials from the user in order to access the network?

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+<3> will manage metadata related to Apps to give uniformity in experience when shifting from one machine to another - eg., if App `A` is installed in machine 1 then when the user logs into machine 2 using his/her SAFE-Account via Launcher, he/she will be presented with a union of all the Apps that were installed on all the machines which access the SAFE-Network on his/her behalf.
+
+<4> Not allow Apps to mutate each other's configs.
+
+<5> Easily Revoke App's ability to Read and Mutate the Network on User's behalf.
+
+# Detailed design
+
+## User's Login Session Packet (for reference)
+This is only to provide a context to the references to it below. This might change in future without affecting this RFC (ie., only a small portion of this is actually relevant for this RFC).
```
Fraser999: "ie.," -> "i.e."

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+```
+Account {
+    an_maid,
+    maid,
+    public_maid,
+    an_mpid,
+    mpid,
+    public_mpid,
+    Option<USER’S-PRIVATE-ROOT-DIRECTORY-ID>, // This is easily accessible to the user (ie., mounted as a drive etc.).
```
Fraser999: If we could stick to a line limit of 100 chars in all code blocks, the rendered view would be unlikely to require horizontal scrolling.

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+    Option<MAIDSAFE-SPECIFIC-CONFIG-ROOT>     // This is accessible only if specifically asked and comes with a warning to not directly modify it.
+}
+```
+- The Root Directories are encrypted with MAID-keys.
+
+## Start Flow
+
+**step 0:** Start Launcher.
+
+**step 1:** Enter Credentials to either create an account or to log into a previously created one.
```
Fraser999: "Credentials" -> "credentials"

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+
+**step 2:** If it was a log in, Launcher Fetches and decodes User-Session-Packet (USP).
```
Fraser999: "Fetches" -> "fetches"

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+
+**step 3:** Launcher Fetches `<MAIDSAFE-SPECIFIC-CONFIG-ROOT>` (See Session Packet description) - if not present Launcher will Create it (via [safe_nfs](https://github.com/maidsafe/safe_nfs) crate)
```
Fraser999: "Fetches" -> "fetches"
"Create" -> "create"

Fraser999: Provide link to Session Packet description:
```
(See [Session Packet description](#users-login-session-packet-for-reference))
```

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+
+**step 4:** Launcher Reads the special Directory reserved for it - `<MAIDSAFE-SPECIFIC-CONFIG-ROOT>/LauncherReservedDirectory`; (See Session Packet Description) - if not present Launcher will Create it (via [safe_nfs](https://github.com/maidsafe/safe_nfs) crate)
```
Fraser999: Reads/Directory/Create to lowercase
Provide link to SP desc.

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+
+**step 5:** Launcher Fetches `<USER’S-PRIVATE-ROOT-DIRECTORY-ID>` - if not present Launcher will Create it (via [safe_nfs](https://github.com/maidsafe/safe_nfs) crate)
```
Fraser999: Fetches/Create to lowercase

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+
+**step 6:** Launcher Checks for special Directory named `SAFEDrive` inside `<USER’S-PRIVATE-ROOT-DIRECTORY-ID>/SAFEDrive` - if not present Launcher will Create it (via [safe_nfs](https://github.com/maidsafe/safe_nfs) crate). This is the directory that the user will usually see it mounted as root.
```
Fraser999: Checks/Directory/Create to lowercase

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+
+**step 7:** Launcher Background Process will listen on local host: `127.0.9.9:30000`. If unavailable, it will increment the port till a valid TCP Listener is in place. This will be thus `127.0.9.9:Launcher-Port`. This will remain on till either the OS is shutdown or Launcher Background Process is killed. We will call this combination `<Laucher-IP:Launcher-Port>`.
```
Fraser999: Did you mean the loopback IP `127.0.0.1`?

Also, this is the first time we've called the Launcher "Launcher Background Process" I think - should we stick with "Launcher" or "SAFE Launcher" throughout?

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+
+## Add App Flow
+
+**step 0:** User Drags `XYZ` App binary into the Launcher to add it.
```
Fraser999: Drags -> drags

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+
+**step 1:** Launcher Creates (if it’s the first time it saw this App):
```
Fraser999: Creates -> creates

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+- Unique Random 64 Byte Id for this app - App-ID (because names of different binaries can be same if they are in different locations on the same machine - thus we need a unique identifier for the binary across all machines)
```
Fraser999: "Random 64 Byte Id" -> "random 64 byte ID"

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+- Unique Directory Id and an associated Unique Root Directory Listing `<APP-ROOT-DIR>` for this app - `XYZ-Root-Dir`. For directory name conflicts append numbers so that they can exist on a file system - eg., `XYZ-1-Root-Dir`. This shall be created inside `<USER’S-PRIVATE-ROOT-DIRECTORY-ID>/`.
```
Fraser999: We haven't explained what a Directory Id or Listing are.  Maybe a brief description or a link to the source code?

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+- `<APP-ROOT-DIR>` shall be always **unversioned** and **private** (ie., encrypted with App-specific crypto keys). Any requirement for a versioned or public directory from the App can be managed by the App itself by creating further subdirectories.
```
Fraser999: > encrypted with App-specific crypto keys
I guess this may change if we move to symmetric encryption - maybe best to just remove it from here?

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+- The user is asked if it's wanted to share the access of SAFEDrive directory with this App.
```
Fraser999: Maybe rephrase to "user is asked if this app should be given read/write access to the `SAFEDrive` directory"?

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+- Append this information in `<LAUNCHER-CONFIG-FILE>` = `<MAIDSAFE-SPECIFIC-CONFIG-ROOT>/LauncherReservedDirectory/LauncherConfigurationFile` (This is what the DNS crate currently does too inside `<MAIDSAFE-SPECIFIC-CONFIG-ROOT>/DnsReservedDirectory/DnsConfigurationFile`.) - Format shall be CBOR (compact-binary-object-representation).
```
Fraser999: "compact" -> "concise" (another instance below too).

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+```
+- Create/Append to local config file (on the user's machine) the following:
+```
+[
+    { App-ID, “/path/to/XYZ” }, // This is newly added by the above
+    { App-ID, “/path/to/some/other/app/already/there” },
+    ... etc.,
+]
+```
+- The format of the config file will be CBOR (compact-binary-object-representation). The name of the local config file should be `<LOCAL-CONFIG-FILE> = launcher.config`. The config file location flowchart shall be same as that of `crust` crate's.
```
Fraser999: Maybe link to docs for this: https://github.com/maidsafe/crust/blob/03d63b4526cf87667c1d1775377398875072c12a/docs/vault_config_file_flowchart.pdf

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+
+**step 2:** User activates the app (eg., double click) from within the Launcher.
```
Fraser999: "eg.," -> "e.g."

proposed/0006-Launcher.md
Fraser999: This local config file could be encrypted with a symmetric key held in the session packet and/or signed with the MAID.

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+
+**step 3:** Launcher checks the App-ID, reads the path from the `<LOCAL-CONFIG-FILE>` that it made and starts the app as an independent process. The Launcher supplies a random port on which it will listen to this app via command line options.
+
+**/path/to/app/binary --launcher "udp:<random-utf8-port>:<random-utf8-string>"**
```
Fraser999: This needs formatted as code so the angle-brackets get rendered.

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+
+All parameters are utf8 Strings.
```
Fraser999: "utf8 Strings" -> "UTF-8 `String`s" or "UTF-8 strings"

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+
+**step 4:** Launcher will wait for a predefined time of **15** seconds for data reception on that port. If it times out it will close the socket (release its binding to it)
+
+**step 5:** App generates a random encryption asymmetric keypair - `<App-Asymm-Keys>`. Then it responds on the socket asking for Launcher to give it an `<App-Specific-Symm-Key>`, its root directory-key and SAFEDrive directory-key, which Launcher had reserved as `XYZ-Root-Dir`
```
Fraser999: "encryption asymmetric" -> "asymmetric encryption"

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+- The payload format for this request shall be CBOR encoded sturcture of the following:
```
Fraser999: "be CBOR" -> "be a CBOR"
"sturcture" -> "structure"

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+    launcher_string      : String, // This must be the one supplied by Launcher
+    nonce                : sodiumoxide::crypto::box_::Nonce,
+    public_encryption_key: sodiumoxide::crypto::box_::PublicKey, // from <App-Asymm-Keys>
+}
+```
+
+**step 6:** The Launcher verifies the `launcher_string` field above and generates a strong random symmetric encryption key `<App-Specific-Symm-Key>`. This is encrypted using App's `public_encrytion_key` and `nonce` above.
+
+**step 7:** Launcher gives the App what it requested.
+- The payload format for this response shall be CBOR encoded sturcture of the following,
```
Fraser999: "be CBOR" -> "be a CBOR"
"sturcture" -> "structure"
"following," -> "following:"

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+```
+struct Response {
+    cipher_text       : Vec<u8>, // encrypted symmetric keys
+    app_root_dir_key  : DirectoryKey,
+    safe_drive_dir_key: Option<DirectoryKey>,
+    tcp_listening_port: u16, // <Laucher-Port>
+}
+
+Where DirectoryKey consists of NameType, u64 (tag), AccessLevel and bool (versioned or not) as defined in safe_nfs crate.
```
Fraser999: Guess this line is not meant to be in the code block?

We could also provide a link to the current source for `DirectoryKey`

proposed/0006-Launcher.md
Fraser999: It should be TCP I think, not UDP?

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+```
+- At this point Launcher closes the UDP socket (ie., is no longer bound to it).
```
Fraser999: UDP -> TCP I think.
"ie.," -> "i.e."

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+
+## Reads and Mutations by the App
+
+**step 0:** App will connect to the Launcher Background process using `<Launcher-IP>:<Launcher-Port>`.
```
Fraser999: "process" -> "Process" if we're using that terminology.

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+
+**step 1:** All `GET/PUT/POST/DELETE`s will go via the Launcher. The App will essentially encrypt using `<App-Specific-Symm-Key>` and pass it to Launcher Background Process which will decrypt and Re-encrypt and sign it using normal process (currently MAID-keys). For `GET`s it will be the reverse - Launcher will eventually encrypt using `<App-Specific-Symm-Key>` before handing the data over to the App.
```
Fraser999: "Re-encrypt" -> "re-encrypt"

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+
+**step 2:** Since `<App-Specific-Symm-Key>` is recognised by Launcher only for current session, there is no security risk and the App will not be able to trick Launcher the next time it starts to use the previous keys to mutate network or read data on its behalf.
```
Fraser999: This isn't really a step - more a note about the process.

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+
+## Share Directory App Flow
+
+Every time the App tries to access `SAFEDrive` Launcher will check the permission in `<MAIDSAFE-SPECIFIC-CONFIG-ROOT>/LauncherReservedDirectory/LauncherConfigurationFile`.
+### Grant Access
```
Fraser999: We could probably combine grant and revoke points into a single one.  Also we should probably specify the level of access the app gets (i.e. R/W).  Something like:

"User can grant the app read/write access to the `SAFEDrive` directory or revoke that access by adding or removing ... respectively"

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+- Delete from `<LOCAL-CONFIG-FILE>` (on the user's machine) the following:
+```
+[
+    { App-ID, “/path/to/XYZ” }, // Remove this, other entries are untouched.
+    ...
+]
+```
+- Remove the SHA512(App-Binary) from the vector in `<LAUNCHER-CONFIG-FILE>`.
+- Decrement `Reference Count` from `<LAUNCHER-CONFIG-FILE>`.
+- If the `Reference Count` is **0** it means that this is the last machine where the App was present. The Launcher shall not delete `<APP-ROOT-DIR>` from within `SAFEDrive` folder. It is user's responsibility to do that as it might contain information (like pictures etc) which the user may not want to lose. Instead the Launcher will ask if the user wants to do that and act accordingly. Similary only after user confirmation will Launcher remove the App entry from the `<LAUNCHER-CONFIG-FILE>`, as it contains necessary metadata to decrypt the App specific directories.
```
Fraser999: "(like pictures etc)" -> "(like pictures, etc.)"
"Similary" -> "Similarly"

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+
+**procedure 1:** While the other procedure would work, but there might be occassions when the user wants to immediately remove the app completely (which also translates as revoke App's permission to mutate network on user's behalf). He may not have access to other machines where the App was installed and maybe currently running and the previous procedure requires him to remove it from all machines. Thus there shall be an option in Launcher to remove App completely irrespective of if it is installed and/or running in other machines. In such cases Launcher will reduce `Reference Count` to 0 and proceed as above for the detection of zero reference count.
```
Fraser999: "procedure would work, but there" -> "procedure would work, there"
Replace  "he", "him", etc. with "user"
"maybe" -> "may be"

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+
+In both procedures, Launcher will terminate the TCP connection with the App and forget the `<App-Specific-Symm-Keys>` so effect of removal and hence revokation is immediate.
```
Fraser999: "revokation" -> "revocation"

proposed/0006-Launcher.md
```diff
@@ -0,0 +1,204 @@
...
+
+## Misc
+- If the App is added to Launcher in one machine, the mention of this will go into `<LAUNCHER-CONFIG-FILE>` as stated previously. It will thus be listed on every machine when user logs into his/her account via Launcher on that machine. However when the App is attempted to be activated on a machine via Launcher where it was not previously added to Launcher then he/she will be prompted to associate a binary. Once done, the information as usual will go into the `<LOCAL-CONFIG-FILE>` on that machine and the user won't be prompted the next time.
+- When the App is started via Launcher, it will first check if the `SHA512(App-Binary)` matches any one of the corresponding entries in the vector in `<LAUNCHER-CONFIG-FILE>`. If it does not Launcher will interpret this as a malacious App that has replaced the App-binary on user's machine, and thus will show a dialog to the user for confirmation  of whether to still continue, because there can be genuine reasons for binary not matching like the App was updated etc.
+
+# Alternatives
+None yet.
```
Fraser999: We have the other Launcher proposal which we should mention here.

proposed/0006-Launcher.md
ustulation: No, I meant `127.0.9.9` - Ports there have much more chances of being free and unused.

proposed/0006-Launcher.md
ustulation: No, it's unlikely to change even if we move to symmetric encryption.

proposed/0006-Launcher.md
ustulation: No, it's UDP - should not require a TCP overhead here

proposed/0006-Launcher.md
ustulation: No, it's UDP

proposed/0006-Launcher.md
Fraser999: OK

proposed/0006-Launcher.md
Fraser999: I think we should stick with the more normal 127.0.0.1 or ::1 unless we see a need to ​*not*​ use it - kinda like benchmarking.  Less surprising for users I think.

proposed/0006-Launcher.md
Fraser999: I think we should use TCP.  It's reliable, whereas if we used UDP we'd have to expect to lose packets or receive them out of sequence.

proposed/0006-Launcher.md
Fraser999: I still think it should be TCP :)
If we ​_do_​ decide to use UDP, I think there are a few places in the RFC where we talk about using TCP sockets/connections.

Viv-Rajkumar:
Probably worth noting that even as a limitation or as an agreed outcome:


> Launcher forces apps to be started from the launcher. This definitely has an impact on UX where users traditionally are more accustomed to app shortcuts in their desktop's / other places which will not work with the current proposal. The same also applies to apps starting with the OS at startup. With just the current proposal, for any app that wishes to access the SAFE network on a user's behalf, the user needs to manually execute it from the launcher


maybe worth documenting any scope of future work to maybe fix/limit this limitation. 

**example** - A partial solution could be something along the lines of providing the launcher process with an arbitrary listener endpoint that shortcuts of apps can invoke and pass the path to the binary they normally would wish to start. At this point have the launcher validate the request by confirming the binary path as valid from its own local config file and if found valid, start the app with the launcher_random_string in its normal approach. Something like this will facilitate a shortcut of an app invoking the launcher instead of its own app to request for its binary get executed by the launcher. I say this is at best a partial solution cos it still needs the launcher initialised and listening before any other app can start or invoke the same request to the launcher listener which might be an issue for OS startup apps

Viv-Rajkumar:
Similarly, even if not a consideration for the current RFC, a local background service process to channel all requests through while maybe ok for desktop platforms, might definitely need a feasibility check on mobile platforms

ustulation:
Points by @Viv-Rajkumar have (now) been noted in the rfc.

proposed/Launcher-Service-Documentation.md
```diff
@@ -0,0 +1,186 @@
...
+### NFS
+- Actions
+```
+"create-dir"
+"delete-dir"
+"create-file"
+"delete-file"
+```
+
+- Create Directory
```
Viv-Rajkumar: Nice start for an initial implementation to the RPCs. Might be worth considering some alternatives to the data section formats for the future too. say for example:
```
{
  endpoint: "safe-api/v1.4/nfs/create-dir"
  data: {
    is_shared: false,
    path: "/a/b/c"
    is_private: true
    ...
}
```

proposed/Launcher-Service-Documentation.md
```diff
@@ -0,0 +1,186 @@
...
+```
+
+### NFS
+- Actions
+```
+"create-dir"
+"delete-dir"
+"create-file"
+"delete-file"
+```
```
Viv-Rajkumar: looks like we're a bit light with anyone wanting to "get" data they've stored :)

PR #43, opened 2015-09-24
MPID Messaging

Fraser999:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/43)

### DIFF
--- a/proposed/MPID Messaging.md
+++ b/proposed/MPID Messaging.md
@@ -219,8 +219,9 @@ enum MpidMessageWrapper {
     Online,
     /// Send out an MpidMessage
     PutMessage(MpidMessage),
-    /// Send out an MpidHeader and sender's original SignedToken to allow response to sender's Put
-    PutHeader(MpidHeader, Option<SignedToken>),
+    /// Send out an MpidHeader and sender's original Authority and SignedToken to allow response
+    /// to sender's Put
+    PutHeader(MpidHeader, Authority, Option<SignedToken>),
     /// Try to retrieve the message corresponding to the included header
     GetMessage(MpidHeader),
     /// List of headers to check for continued existence of corresponding messages in Sender's outbox
@@ -410,10 +411,6 @@ Pseudo-code for MpidManager:
 pub struct MpidManager {
     outbox_storage: Vec<OutboxAccount>,
     inbox_storage: Vec<InboxAccount>,
-    on_going_puts: lru_time_cache<message_name: NameType, (sender_client: ::routing::Authority::Client,
-                                                           token: Option<::routing::SignedToken>)>,
-    on_going_gets: lru_time_cache<message_name: NameType, (recipient_client: ::routing::Authority::Client,
-                                                           token: Option<::routing::SignedToken>)>,
     routing: Routing,
 }
 
@@ -425,7 +422,6 @@ impl MpidManager {
         if messaging {  // sd.data holds MpidMessage
             // insert received mpid_message into the outbox_storage
             if outbox_storage.insert(from, mpid_message) {
-                on_going_puts.insert(mpid_message_name(mpid_message), (from, token));
                 let forward_sd = StructuredData {
                     type_tag: MPID_MESSAGE,
                     identifier: mpid_message_name(mpid_message),
@@ -468,7 +464,6 @@ impl MpidManager {
             let recipient_account = inbox_storage.find_account(to);
             if recipient_account.recipient_clients.len() > 0 { // indicates there is connected client
                 for header in recipient_account.headers {
-                    on_going_gets.insert(mpid_header_name(header.mpid_header), (from, token));
                     get_message(header);
                 }
             }
@@ -480,7 +475,8 @@ impl MpidManager {
     //     2, remove_header: delete request from recipient B to MpidManagers(B)
     pub fn handle_delete(from, to, name) {
         if remove_message {  // from.name != to.name
-            remove the message (bearing the name) from sender(to.name)'s account if the message's specified recipient is the requester (from);
+            remove the message (bearing the name) from sender(to.name)'s account if the message's
+            specified recipient is the requester (from);
         }
         if remove_header {  // from.name == to.name
             remove the header (bearing the name) from recipient(from.name)'s account;
@@ -505,14 +501,14 @@ impl MpidManager {
         if replying {  // MpidManager(A) replies to MpidManager(B) with the requested mpid_message
             let account = inbox.find_account(to_name);
             if account.has_header(mpid_message.name()) {
-                let (reply_to, token) = on_going_gets.find(mpid_message.name());
-                foward the mpid_message to client via routing.post using (reply_to, token);
+                forward the mpid_message to client via routing.post using (reply_to, token);
             }
         }
         if fetching {
             if outbox.has_message(name) {
                 // recipient's MpidManager asking for a particular message and it exists
-                if the requester is the recipient, reply message to the requester(from) with outbox.find_message(name) via routing.post;
+                if the requester is the recipient, reply message to the requester(from) with
+                outbox.find_message(name) via routing.post;
             } else {
                 // recipient's MpidManager asking for a particular message but not exists
                 reply failure to the requester(from);
@@ -521,19 +517,18 @@ impl MpidManager {
     }
 
     // handle_post_response:
-    //     1, no_record: response contains Error msg holding the original get request which has ori_mpid_header_name
-    //     2, inbox_full: response contains Error msg holding the original put request which has ori_mpid_header
+    //     1, no_record: response contains Error msg holding the original get request which has
+    //        original_mpid_header_name
+    //     2, inbox_full: response contains Error msg holding the original put request which has
+    //        original_mpid_header
     pub fn handle_post_failure(from, to, response) {
-        if no_record {  // MpidManager(A) replies to MpidManager(B) that the requested mpid_message doesn't exists
-            remove the header (bearing the ori_mpid_header_name) from the account of to.name;
-            if on_going_gets.has(ori_mpid_header_name) {
-                let (reply_to, token) = on_going_gets.find(ori_mpid_header_name);
-                send failure to client via routing.get_failure using (reply_to, token, mpid_header);
-            }
+        if no_record {
+            // MpidManager(A) replies to MpidManager(B) that the requested mpid_message doesn't exist
+            remove the header (bearing the original_mpid_header_name) from the account of to.name;
         }
         if inbox_full {  // MpidManager(B) replies to MpidManager(A) that inbox is full
-            remove the message (bearing the ori_mpid_header.name()) from the account of to.name;
-            let (reply_to, token) = on_going_puts.find(ori_mpid_header.name());
+            remove the message (bearing the original_mpid_header.name()) from the account of to.name;
+            original sender's `reply_to` and `token` will be available in this incoming message
             send failure to client via routing.put_failure using (reply_to, token, message);
         }
     }
--- a/proposed/MPID Messaging.md
+++ b/proposed/MPID Messaging.md
@@ -247,17 +247,17 @@ Requests composed by Client:
 
 Requests composed by Client:
 
-| Request      | Usage Scenario                                             | Content                                              | From Authority | Destination Authority |
-|:-------------|:-----------------------------------------------------------|:-----------------------------------------------------|:---------------|:----------------------|
-| PutResponse  | put failure (outbox or inbox full)                         | `Error(Wrapper::MpidMessage)`                        | Managers(A)    | Client(A)             |
-| Put          | new message notification                                   | `Wrapper::MpidHeader`                                | Managers(A)    | Managers(B)           |
-| PutResponse  | put failure (inbox full)                                   | `Error(Wrapper::MpidHeader)`                         | Managers(B)    | Managers(A)           |
-| Post         | retrieve message from sender's outbox to pass to recipient | `Wrapper::GetMessage(Header.name)`                   | Managers(B)    | Managers(A)           |
-| PostResponse | requested message no longer available in sender's outbox   | `Error(Wrapper::GetMessage(Header.name))`            | Managers(A)    | Managers(B)           |
-| Post         | pass message from sender's outbox to recipient's managers  | `Wrapper::MpidMessage`                               | Managers(A)    | Managers(B)           |
-| Post         | push message to intended recipient                         | `Wrapper::MpidMessage`                               | Managers(B)    | Client(B)             |
-| Post         | reply to client's `OutboxHas` request                      | `Wrapper::OutboxHasResponse(Vec<MpidHeader>)`        | Managers(A)    | Client(A)             |
-| Post         | reply to client's `GetOutboxHeaders` request               | `Wrapper::GetOutboxHeadersResponse(Vec<MpidHeader>)` | Managers(A)    | Client(A)             |
+| Request      | Usage Scenario                                             | Content                                                 | From Authority | Destination Authority |
+|:-------------|:-----------------------------------------------------------|:--------------------------------------------------------|:---------------|:----------------------|
+| PutResponse  | put failure (outbox or inbox full)                         | `Error(Wrapper::MpidMessage)`                           | Managers(A)    | Client(A)             |
+| Put          | new message notification                                   | `Wrapper::MpidHeader`                                   | Managers(A)    | Managers(B)           |
+| PutResponse  | put failure (inbox full)                                   | `Error(Wrapper::MpidHeader)`                            | Managers(B)    | Managers(A)           |
+| Post         | retrieve message from sender's outbox to pass to recipient | `Wrapper::GetMessage(Header.name)`                      | Managers(B)    | Managers(A)           |
+| PostResponse | requested message no longer available in sender's outbox   | `Error(Wrapper::GetMessage(` `Header.name))`            | Managers(A)    | Managers(B)           |
+| Post         | pass message from sender's outbox to recipient's managers  | `Wrapper::MpidMessage`                                  | Managers(A)    | Managers(B)           |
+| Post         | push message to intended recipient                         | `Wrapper::MpidMessage`                                  | Managers(B)    | Client(B)             |
+| Post         | reply to client's `OutboxHas` request                      | `Wrapper::OutboxHasResponse(` `Vec<MpidHeader>)`        | Managers(A)    | Client(A)             |
+| Post         | reply to client's `GetOutboxHeaders` request               | `Wrapper::GetOutboxHeadersResponse(` `Vec<MpidHeader>)` | Managers(A)    | Client(A)             |
 
 MPID Header:
 
--- a/proposed/MPID Messaging.md
+++ b/proposed/MPID Messaging.md
@@ -170,19 +170,6 @@ We also have identified a need for some form of secure messaging in order to imp
 
 # Unresolved Questions
 
-1. `MpidMessage` and `MpidHeader` are wrapped in `StructuredData` instances in order to allow Delete requests on them and to allow them to be given different tags (otherwise they could be ImmutableData, since they don't mutate).  Should Routing, which already has to define these two types, be made "aware" of these two types?  (i.e. should they get added to the [`::routing::data::Data` enum][6])
-
-    Identified pros:
-    - less wrapping/parsing, so simpler for Vaults and Clients to deal with
-    - more efficient (smaller messages to transmit and store)
-    - no need for Routing to publicly expose `MPID_MESSAGE_TAG` or `MPID_HEADER_TAG`
-
-    Identified cons:
-    - increased complexity in Routing
-    - would need to add the sender's public key to the header Put flow
-
-    Qi prefers using `StructuredData`, Fraser prefers not using `StructuredData` unless the effort required by Routing to accommodate the new types is significant.
-
 # Future Work
 
 - This RFC doesn't address how Clients get to "know about" each other.  Future work will include details of how Clients can exchange MPIDs in order to build an address book of peers.
@@ -205,7 +192,7 @@ We also have identified a need for some form of secure messaging in order to imp
 
 All MPID-related messages will be in the form of a Put, Post or Delete of a `StructuredData`.
 
-This will be:
+Such `StructuredData` will be:
 
 ```rust
 
@@ -221,35 +208,6 @@ StructuredData {
 
 ```
 
-where XXX is indicated in the following list of all MPID-related messages:
-
-Sent from Client:
-
-| Message | From ==> To |
-|:---|:---|
-| Post[Online] |           Either Client    ==> Own Managers |
-| Put[Message] |           Sender Client    ==> Own Managers |
-| Post[Has[Vec<Header>]] | Sender Client    ==> Own Managers |
-| Post[GetAllHeaders] |    Sender Client    ==> Own Managers |
-| Delete[Header] |         Either Client    ==> Own Managers (try to delete from inbox and outbox) |
-| Delete[Header] |         Recipient Client ==> Sender's Managers |
-
-
-Sent from Vault:
-
-| Message | From ==> To |
-|:---|:---|
-| PutResponse[Error[Message]] |              Sender's Managers (outbox)   ==> Sender Client |
-| Put[Header] |                              Sender's Managers (outbox)   ==> Recipient's Managers (inbox) |
-| PutResponse[Error[Header]] |               Recipient's Managers (inbox) ==> Sender's Managers (outbox) |
-| Post[GetMessage[Header]] |                 Recipient's Managers (inbox) ==> Sender's Managers (outbox) |
-| PostResponse[Error[GetMessage[Header]]] |  Sender's Managers (outbox)   ==> Recipient's Managers (inbox) |
-| Post[Message] |                            Sender's Managers (outbox)   ==> Recipient's Managers (inbox) |
-| Post[Message] |                            Recipient's Managers (inbox) ==> Recipient Client |
-| Post[HasResponse[Vec<Header>]] |           Sender's Managers (outbox)   ==> Sender Client |
-| Post[GetAllHeadersResponse[Vec<Header>]] | Sender's Managers (outbox)   ==> Sender Client |
-
-
 The various different types for `StructuredData::data` can be enumerated as:
 
 ```rust
@@ -257,19 +215,49 @@ The various different types for `StructuredData::data` can be enumerated as:
 enum MpidMessageWrapper {
     /// Notification that the MPID Client has just connected to the network
     Online,
+    /// Send out an MpidMessage
+    PutMessage(MpidMessage),
+    /// Send out an MpidHeader and sender's original SignedToken to allow response to sender's Put
+    PutHeader(MpidHeader, Option<SignedToken>),
     /// Try to retrieve the message corresponding to the included header
     GetMessage(MpidHeader),
     /// List of headers to check for continued existence of corresponding messages in Sender's outbox
-    Has(Vec<MpidHeader>),
+    OutboxHas(Vec<MpidHeader.name()>),
     /// Subset of list from Has request which still exist in Sender's outbox
-    HasResponse(Vec<MpidHeader>),
+    OutboxHasResponse(Vec<MpidHeader>),
     /// Retrieve the list of headers of all messages in Sender's outbox
-    GetAllHeaders,
+    GetOutboxHeaders,
     /// The list of headers of all messages in Sender's outbox
-    GetAllHeadersResponse(Vec<MpidHeader>),
+    GetOutboxHeadersResponse(Vec<MpidHeader>),
 }
 ```
 
+The following list of all MPID-related messages show how this enum is used.  (In these tables, Client(A) is the original sender, Client(B) is the recipient, and Managers(A) and Managers(B) are their respective MpidManagers ).
+
+Requests composed by Client:
+
+| Request | Usage Scenario                                                      | Content                                | Destination Authority           |
+|:--------|:--------------------------------------------------------------------|:---------------------------------------|:--------------------------------|
+| Put     | Client(A) creating a new message                                    | `Wrapper::MpidMessage`                 | Managers(A)                     |
+| Post    | Client(A) checking existence of list of sent messages in own outbox | `Wrapper::OutboxHas(Vec<Header.name>)` | Managers(A)                     |
+| Post    | Client(A) getting list of all messages still in own outbox          | `Wrapper::GetOutboxHeaders`            | Managers(A)                     |
+| Delete  | Client(A) or (B) deleting from own outbox or inbox respectively     | `NameType`                             | Managers(A) or (B) respectively |
+| Delete  | Client(B) deleting a "read" message from sender's outbox            | `NameType`                             | Managers(A)                     |
+| Post    | Client announcing to Managers it's connected to network             | `Wrapper::Online`                      | Managers                        |
+
+Requests composed by Client:
+
+| Request      | Usage Scenario                                             | Content                                              | From Authority | Destination Authority |
+|:-------------|:-----------------------------------------------------------|:-----------------------------------------------------|:---------------|:----------------------|
+| PutResponse  | put failure (outbox or inbox full)                         | `Error(Wrapper::MpidMessage)`                        | Managers(A)    | Client(A)             |
+| Put          | new message notification                                   | `Wrapper::MpidHeader`                                | Managers(A)    | Managers(B)           |
+| PutResponse  | put failure (inbox full)                                   | `Error(Wrapper::MpidHeader)`                         | Managers(B)    | Managers(A)           |
+| Post         | retrieve message from sender's outbox to pass to recipient | `Wrapper::GetMessage(Header.name)`                   | Managers(B)    | Managers(A)           |
+| PostResponse | requested message no longer available in sender's outbox   | `Error(Wrapper::GetMessage(Header.name))`            | Managers(A)    | Managers(B)           |
+| Post         | pass message from sender's outbox to recipient's managers  | `Wrapper::MpidMessage`                               | Managers(A)    | Managers(B)           |
+| Post         | push message to intended recipient                         | `Wrapper::MpidMessage`                               | Managers(B)    | Client(B)             |
+| Post         | reply to client's `OutboxHas` request                      | `Wrapper::OutboxHasResponse(Vec<MpidHeader>)`        | Managers(A)    | Client(A)             |
+| Post         | reply to client's `GetOutboxHeaders` request               | `Wrapper::GetOutboxHeadersResponse(Vec<MpidHeader>)` | Managers(A)    | Client(A)             |
 
 MPID Header:
 
--- a/proposed/MPID Messaging.md
+++ b/proposed/MPID Messaging.md
@@ -459,11 +459,11 @@ impl MpidManager {
 
     // get messages or headers on request:
     pub fn handle_get(from, to, name, token) {
-        if out_box.has_account(name) {
+        if outbox.has_account(name) {
             // sender asking for the headers of existing messages
-            reply to the requester(from) with out_box.find_account(name).get_headers() via routing.post;
+            reply to the requester(from) with outbox.find_account(name).get_headers() via routing.post;
         }
-        if in_box.has_account(name) {
+        if inbox.has_account(name) {
             // triggering pushing all existing messages to client, first needs to fetch them
             let recipient_account = inbox_storage.find_account(to);
             if recipient_account.recipient_clients.len() > 0 { // indicates there is connected client
@@ -475,7 +475,7 @@ impl MpidManager {
         }
     }
 
-    // revoming message or header on request:
+    // removing message or header on request:
     //     1, remove_message: delete request from recipient B to sender's MpidManagers(A)
     //     2, remove_header: delete request from recipient B to MpidManagers(B)
     pub fn handle_delete(from, to, name) {
@@ -500,7 +500,7 @@ impl MpidManager {
                 send a get request to the sender's MpidManager asking for the full message;
             }
             let mut sender_account = outbox.find_account(mpid_name);
-            sender_account.register_on_line(mpid_client);
+            sender_account.register_online(mpid_client);
         }
         if replying {  // MpidManager(A) replies to MpidManager(B) with the requested mpid_message
             let account = inbox.find_account(to_name);
@@ -510,9 +510,9 @@ impl MpidManager {
             }
         }
         if fetching {
-            if out_box.has_message(name) {
+            if outbox.has_message(name) {
                 // recipient's MpidManager asking for a particular message and it exists
-                if the requester is the recipient, reply message to the requester(from) with out_box.find_message(name) via routing.post;
+                if the requester is the recipient, reply message to the requester(from) with outbox.find_message(name) via routing.post;
             } else {
                 // recipient's MpidManager asking for a particular message but not exists
                 reply failure to the requester(from);
@@ -577,7 +577,7 @@ pub fn send_mpid_message(my_mpid: Mpid, recipient: ::routing::Authority::Client,
                                ::routing::data::Data::StructuredData(sd));
 }
 
-/// Client register to be on-line
+/// Client register to be online
 pub fn register_online(my_mpid: Mpid) {
     let sd = StructuredData {
         type_tag: MPID_MESSAGE,
@@ -633,7 +633,7 @@ General functions
 
 ```rust
 pub fn mpid_header_name(mpid_header: &MpidHeader) -> ::routing::NameType {
-    ::crypto::hash::sha512::hash(::util::encode(mpid_header))
+    ::crypto::hash::sha512::hash(::utils::encode(mpid_header))
 }
 pub fn mpid_message_name(mpid_message: &MpidMessage) -> ::routing::NameType {
     mpid_header_name(mpid_message.mpid_header)


proposed/0008 MPID Messaging.md
```diff
@@ -0,0 +1,295 @@
...
+    - more efficient (smaller messages to transmit and store)
+    - no need for Routing to publicly expose `MPID_MESSAGE_TAG` or `MPID_HEADER_TAG`
+
+    Identified cons:
+    - increased complexity in Routing
+    - would need to add the sender's public key to the header Put flow
+
+    Qi prefers using `StructuredData`, Fraser prefers not using `StructuredData` unless the effort required by Routing to accommodate the new types is significant.
+
+1. By having the `recipient` field inside `MpidMessage::signed_header`, a `sign::verify` call is required at every hop to parse the header and identify the recipient.  An alternative is to duplicate the `recipient` field in the `MpidMessage` itself.
```
maqi: An alternative is to put the `recipient` field in the `MpidMessage` instead.  (not a duplication but a move)
this will eliminates the concern of increased message size and need to carry out consistency checks.
The only identified con is: `recipient` is not protected by signed_header
However this will be easily overcome by have signed_body = sign(recipient+content)
I prefer to move the `recipient` from `MpidHeader` to `MpidMessage`.

proposed/MPID Messaging.md
```diff
@@ -0,0 +1,295 @@
...
+    - more efficient (smaller messages to transmit and store)
+    - no need for Routing to publicly expose `MPID_MESSAGE_TAG` or `MPID_HEADER_TAG`
+
+    Identified cons:
+    - increased complexity in Routing
+    - would need to add the sender's public key to the header Put flow
+
+    Qi prefers using `StructuredData`, Fraser prefers not using `StructuredData` unless the effort required by Routing to accommodate the new types is significant.
+
+1. By having the `recipient` field inside `MpidMessage::signed_header`, a `sign::verify` call is required at every hop to parse the header and identify the recipient.  An alternative is to duplicate the `recipient` field in the `MpidMessage` itself.
```
maqi: An alternative is to put the recipient field in the MpidMessage instead. (not a duplication but a move)
this will eliminates the concern of increased message size and need to carry out consistency checks.
The only identified con is: recipient is not protected by signed_header
However this will be easily overcome by have signed_body = sign(recipient+content)
I prefer to move the recipient from MpidHeader to MpidMessage.

proposed/MPID Messaging.md
```diff
@@ -0,0 +1,295 @@
...
+    1. detecting of Client, if "push" model used
+    1. MPID Client addressing (if MPID address registration procedure is to be undertaken - i.e. in "pull" model)
+
+1. Routing
+    1. `Authority::MpidManager`
+    1. definition of `MPID_MESSAGE_TAG` and `MPID_HEADER_TAG`
+    1. definition of `MpidMessage` and `MpidHeader`
+    1. support Delete (for StructuredData only)
+    1. if "push" model: notify vault when a client joins and notifying Client with `MpidHeader`
+    1. address relocation (allows Client to use a fixed MPID address to connect to the network and allows "push" model - not required for "pull" model)
```
maqi: Checked with the current Address Relocation proposal (https://github.com/maidsafe/rfcs/pull/40).
The capability of using fixed address for client to connect to network is not included. (connecting address is always dynamic calculated based on the close group, which is not determined)
This leave Mpid Messaging can only adopt Pull model or using Address Registration approach.

Fraser999: Well, the pull model was our preference anyway, so I guess we can update the RFC to target the pull model, and maybe just mention the push model as future work?

proposed/MPID Messaging.md
Fraser999: So this means we have a detached signature in `MpidMessage` or the `recipient` is duplicated in the `MpidMessage`.  So the new structs would be:

```rust
pub struct MpidHeader {
    pub sender: ::routing::NameType,
    pub index: u32,
    pub parent_index: Option<u32>,
    pub subject: Vec<u8>,
}

pub struct MpidMessage {
    pub signed_header: Vec<u8>,
    pub recipient: ::routing::NameType,
    pub body: Vec<u8>,
    pub signature_of_recipient_and_body: ::sodiumoxide::crypto::sign::Signature,
}
```

I prefer this to duplicating the `recipient` field, but I still favour the original approach.  I don't feel there's any real harm in forcing every node to validate the signature at every step just to parse the header.

Even if we do find with benchmarking that this is an issue, we can subvert the system by ignoring the signature part (which is always a fixed size, and is always prepended) of the `signed_header` and parsing the remaining part.

proposed/MPID Messaging.md
maqi: I agree with the idea of updating the RFC to target the pull model and mention the push model as future work  :)

proposed/MPID Messaging.md
maqi:
```rust
pub struct MpidHeader {
    pub sender: ::routing::NameType,
    pub index: u32,
    pub parent_index: Option<u32>,
    pub subject: Vec<u8>,
}
pub struct ContentAndRecipient {
    pubrecipient: Vec<u8>,
    pub body: Vec<u8>,
}
pub struct MpidMessage {
    pub signed_header: Vec<u8>,
    pub signed_recipient_and_body: Vec<u8>,
}
```
Actually, if non-detached signature being used, `recipient` doesn't need to be explicitly presented in MpidMessage at all. And the enforcement of signature checking still applies within the above proposal as recipient needs to be verified and parsed.

Fraser999: OK - but now I don't see how this better than the original proposal?  I thought your point was that neither MpidManagers(A) nor MpidManager(B) *need* to check that A has correctly signed.  Routing guarantees that the message has come from Client(A), and if A has put a bad signature, Client(B) will catch it.  Or if one set of the MpidManagers does the verification, there's no need for the other set to also do it.

This latest suggestion still requires every hop to check the signature (or "cheat" by ignoring the signature portion of the `Vec<u8>`) doesn't it?

I still prefer the original suggestion :)

proposed/MPID Messaging.md
```diff
@@ -0,0 +1,295 @@
...
+
+## Planned Work
+
+1. Vault
+    1. outbox
+    1. inbox
+    1. sending message flow
+    1. retrieving message flow
+    1. deleting message flow
+    1. churn handling and refreshing for account_transfer
```
maqi: When churn happens, the account entry to be transferred among MpidManagers can be MpidMessage (for Outbox) or MpidHeader (for inbox), i.e. one MpidMessage or MpidHeader per transferring entry.
`Grouping` the entries, i.e. one Outbox or Inbox pmid_account per transferring entry, defines an account transfer entry as :
```rust
struct MpidMessageAccount {
    pub sender: ::routing::NameType,
    pub mpid_messages: Vec<MpidMessage>,
}
struct MpidHeaderAccount {
    pub recipient: ::routing::NameType,
    pub senders: Vec<::routing::NameType>,
    pub mpid_headers: Vec<signed_header: Vec<u8>>,
}
```
may cause the transferring account entry being too huge.

proposed/MPID Messaging.md
maqi: The benefit of my latest proposal compared with the original one is :  
1, the storage of Inbox can be reduced
2, the message size among MpidManagers and from MpidManager to recipient can be reduced.

If `every hop verification` is preferred, my latest proposal still fits, otherwise, your proposal above will be enough.

Fraser999: Yes - OK - I agree.  I think my only niggle would now be that the new struct be named `RecipientAndBody`.

proposed/MPID Messaging.md
Fraser999: Maybe we should aim to group the entries, but for now limit the size of the outbox and inbox so that we know we can handle them safely with churn?

maqi: It seems there will be issue once the message size becomes greater than couple of MBytes.
Meanwhile, I think limit an outbox size capped at just couple of MBytes is problematic.  :(
Maybe we can have MpidHeader to be grouped and leave MpidMessage not grouped ?

Fraser999: Yeah, I agree if the message size is limited to just 2MB.  I know we had that limitation in the C++ code - I didn't know we still had it.

proposed/MPID Messaging.md
maqi: Although `Pull` model is preferred here, according to the comments in RFC https://github.com/maidsafe/rfcs/pull/40, the churn_client (notification of client joining/leaving) may need to be implemented anyway.
So `Push` model may not incur much complexity as presents in this RFC.

proposed/MPID Messaging.md
```diff
@@ -0,0 +1,295 @@
...
+    pub recipient: ::routing::NameType,
+    pub index: u32,
+    pub parent_index: Option<u32>,
+    pub subject: Vec<u8>,
+}
+```
+
+The `sender` and `recipient` fields are hopefully self-explanatory.  However, the only purpose for them existing in the `MpidHeader` is to increase the randomness of the derived message name so conflicts have a much lower chance of happening.
+
+The `index` and `parent_index` are intended to allow threading or sequencing of messages, while `subject` allows passing arbitrary data (e.g. a subject line).
```
ustulation: While `sequencing` is sort of guessable, I am not completely sure what `threading` means here. May be a little more explanation on this would be helpful.

proposed/MPID Messaging.md
```diff
@@ -0,0 +1,295 @@
...
+
+The `subject` field must not exceed `MAX_SUBJECT_SIZE` bytes.
+
+### `MpidMessage`
+
+```rust
+pub struct MpidMessage {
```
ustulation: Why does this contain `signed` stuffs, i.e. why is signing important here ? Are these are meant to support updation once sent (i.e clients being able to modify sent messages in their outbox) ? If no, then isn't the signing provided by Routing on the entire message enough ?

proposed/MPID Messaging.md
```diff
@@ -0,0 +1,295 @@
...
+```
+                    MpidManagers(A)    MpidManagers(B)
+                   /      *                   *        \
+MpidClient(A) ->  -       *                   *         -  <- MpidClient(B)
+                   \      *                   *        /
+
+```
+1. `MpidClient(A)` sends an `MpidMessage` to `MpidManagers(A)` as a Put request.
+1. `MpidManagers(A)` store this message in their outbox for A and send the `signed_header` component to `MpidManagers(B)` as a Put request, unless the outbox is full in which case a PutFailure response is returned to `MpidClient(A)`.
+1. `MpidManagers(B)` try to store the `(sender, signed_header)` in their inbox for B.  If successful, they forward it to `MpidClient(B)` as a Put request immediately or as soon as it appears online.  If unsuccessful (e.g. the inbox is full or sender has been blacklisted), they reply to `MpidManagers(A)` with a PutFailure, who then remove the message from the outbox and send a PutFailure response to `MpidClient(A)`.
```
ustulation: Do these steps assume that there has been a previous exchange of `ownership` and `encryption` keys between the clients ? If so it might be worth showing how that precursory series of steps happened since that would have been the start of communication, not this messaging.

proposed/MPID Messaging.md
```diff
@@ -0,0 +1,295 @@
...
+1. `MpidManagers(A)` send a GetResponse to `MPidClient(B)`.
+1. On receiving the message, `MpidClient(B)` sends a remove request to `MpidManagers(B)` via a Delete.
+1. `MpidManagers(B)` remove the corresponding entry from the inbox for B and forward the remove request to `MpidManagers(A)`.  `MpidManagers(A)` then remove the corresponding entry from the outbox for A.
+
+`MpidClient(A)` might decide to remove the `MpidMessage` from the outbox if the message hasn't been retrieved by `MpidClient(B)` yet.  In this case, `MpidManagers(A)` should not only remove the corresponding `MpidMessage` from their outbox for A, but also send a notification to the group of `MpidManagers(B)` so they can remove the corresponding entry from their inbox of B.  These messages will all be Delete requests.
+
+_MPidClient(A)_ =>> |__MPidManagers(A)__ (Put)(Header.So) *->> | __MPidManagers(B)__  (Store(Header))(Online(MpidClient(B)) ? Header.So : (WaitForOnlineB)(header.So)) *-> | _MpidClient(B)_ So.Retrieve ->> | __MpidManagers(A)__ So.Message *-> | _MpidClient(B)_ Remove.So ->> | __MpidManagers(B)__ {Remove(Header), Remove.So} *->> | __MpidManagers(A)__ Remove
```
ustulation: Is this some well understood format, because otherwise it's not a very comfortable read.

proposed/MPID Messaging.md
```diff
@@ -0,0 +1,295 @@
...
+The `index` and `parent_index` are intended to allow threading or sequencing of messages, while `subject` allows passing arbitrary data (e.g. a subject line).
+
+The `subject` field must not exceed `MAX_SUBJECT_SIZE` bytes.
+
+### `MpidMessage`
+
+```rust
+pub struct MpidMessage {
+    pub signed_header: Vec<u8>,
+    pub signed_body: Vec<u8>,
```
ustulation: A structured data will have this fragmented a lot if it happens to be > 100KiB (which is very common if there are images/attachments etc.). The `index` doesn't seem to be of a particular use here for reassembly. Has this been given thought ?

proposed/MPID Messaging.md
```diff
@@ -0,0 +1,295 @@
...
+                    MpidManagers(A)    MpidManagers(B)
+                   /      *                   *        \
+MpidClient(A) ->  -       *                   *         -  <- MpidClient(B)
+                   \      *                   *        /
+
+```
+1. `MpidClient(A)` sends an `MpidMessage` to `MpidManagers(A)` as a Put request.
+1. `MpidManagers(A)` store this message in their outbox for A and send the `signed_header` component to `MpidManagers(B)` as a Put request, unless the outbox is full in which case a PutFailure response is returned to `MpidClient(A)`.
+1. `MpidManagers(B)` try to store the `(sender, signed_header)` in their inbox for B.  If successful, they forward it to `MpidClient(B)` as a Put request immediately or as soon as it appears online.  If unsuccessful (e.g. the inbox is full or sender has been blacklisted), they reply to `MpidManagers(A)` with a PutFailure, who then remove the message from the outbox and send a PutFailure response to `MpidClient(A)`.
+1. To retrieve the message from the network, `MpidClient(B)` sends a Get request to `MpidManagers(A)`.
```
ustulation: A Few questions here:
1. If the steps are mandatory for the clients to perform, it is better to do it at the Vault level to make it more deterministic, i.e. when MpidManagers(A) receive the `GET` request from `Client(B)`, they should do both the steps - give data to the client **and** initiate purge stuff from MpidManagers(B) and self. That way it is closer to being Network-atomic than if the clients are involved. This is consequent to the notion that Clients are temporary entities on the Network while Vaults are statistically permanent thus operations initiated by them are more deterministic.
2. Inbox never seems to get the actual content. If there is a need for data to be stored in Inbox even after the Client has retrieved it, then with this approach the sender will have to sacrifice his storage (since it is only there that data is maintained). Could it be more useful if `B` could retrieve messages from `MpidManagers(A)` into `MpidManagers(B)` (i.e. move them) during `GET`s so that the messages could be stored there till `B` wants them deleted ?

proposed/MPID Messaging.md
maqi: I prefer use timestamp as mentioned in the Unresolved Questions

proposed/MPID Messaging.md
maqi: routing signing only ensures a security between a routing peer.
the signed stuff here is to ensure a security between a messaging peer (on vault level).
messages will be stored in OutBox and Inbox, once it is restored and forwarded again, the content is open to be modified and routing sign can not eliminate that risk.

proposed/MPID Messaging.md
maqi: there is no need to exchange `ownership` or `encryption key` for messaging purpose.
sender always holds own priv key, and recipient always know sender's pub-key as the name itself is it.

proposed/MPID Messaging.md
maqi: it is internal format. probably remove this as the doc is a public rfc

proposed/MPID Messaging.md
maqi: what's the current limited size for `data` filed in StructuredData ?

proposed/MPID Messaging.md
maqi: 1, MpidManager can ensure whether client has really received the data or not. If MpidManager send the data to client, but client didn't received it for some reason (say connection disruption) and asking for the msg again, then `purge` started by MpidManager will be problematic.
2, `OutBox` and `Inbox` concepts used here are all on `MpidManagers` as temporarily for transferring purpose. Once a client `received` the message, the message itself shall be moved to its local storage or its own account's online storage. so `received` here is more like `agreed to use my own storage to accept`.

proposed/MPID Messaging.md
Fraser999: I agree with Qi that these would be better being replaced by a single timestamp field.  However, I also take your point Spandan that we don't explain this well enough.

If these fields remain in the submitted RFC, I'll add something like the following:

`index` and `parent_index` can be set to any arbitrary values the sender wishes - they are not used by the network other than to help disambiguate messages.  From the client's perspective, the intent is that they can be used to help create "threaded" conversations between users, i.e. a message can be the first of a new thread by setting `parent_index` to `None`, or it can be a reply to message with `index` x by setting `parent_index` to `Some(x)`.

proposed/MPID Messaging.md
Fraser999: It also allows the message to be fully validatable outside of the context of the SAFE network, i.e. if you have a copy of message from user X, it contains all the information required to validate the message without querying the network.

proposed/MPID Messaging.md
Fraser999: I think Spandan is correct in that there does have to be a preliminary exchange of MPID IDs and public keys though, and we haven't specified how that is to happen.

I think the basic idea is that two human users need to exchange their MPID IDs outside of the SAFE Network (e.g. email them to eachother).  The IDs should be hex-/base32-/base64-encoded whenever they're presented to the user or taken from the user.

The clients will then have to retrieve the corresponding MPID public keys from the network and presumably hold these in some form of address book for the user.

proposed/MPID Messaging.md
Fraser999: I've been meaning to get round to adding a flowchart for the message flow, since this may be easier to read than either of the two formats above?  What do you think @ustulation ?

proposed/MPID Messaging.md
ustulation: There is no size for `data` in particular. However, the size of SD itself must be < 100 KiB to be valid.

Fraser999: > The `index` doesn't seem to be of a particular use here for reassembly.

Yeah - it's not meant for that, but hopefully my comment on `index`/`parent_index` makes that clear now? 

> Has this been given thought ?

I hadn't remembered that - thanks for the reminder :)

I think this lends more weight to my preference for Routing handling these new types directly, avoiding the need to wrap them in `StructuredData` instances.

proposed/MPID Messaging.md
ustulation: Yes @Fraser999 , that would be way better i think

Fraser999: OK - I'll have a stab at that then.

proposed/MPID Messaging.md
ustulation: @Fraser999 : yes now i understand a little more.

@maqi :
> sender's name itself is the pub-key
It is the SHA512 of PublicKey so PublicKey is not derivable from sender's name. And since (unless i overlooked) nothing is mentioned about it, i assume it `SHA512(MAID-Public-signing-Key)`. If so this has all the more nothing to do with MPID-keys. If not then i think it should be mentioned what exactly is this name (or if it already is and i overlooked then pls give me a hint).

proposed/MPID Messaging.md
ustulation: Ok .. It makes sense in combination with other answer:
> The clients will then have to retrieve the corresponding MPID public keys from the network and presumably hold these in some form of address book for the user

So will this be the verification flow ? :
1. Receiver Client (`B`) stores in address-book mapping of `sender name -> sender MPID Public Sign Key`
1. Since from the discussions, it seems `pull` model is heavily favored, `B` does a `GET` for the messages in Inbox and gets `(sender name <-> signed, also probably encrypted, message_header)`.
1. `B` obtains `sender MPID Public Sign Key` from its address book and verifies the received message_header, similarly proceeds to `GET` message_body.
1. If not present in `B`'s address-book then there is no way B can verify the signature and then it's `B`'s call whether to act on it or discard it.

Is that right ? ^^^

proposed/MPID Messaging.md
maqi: @Fraser999 @ustulation : oh, I always have the impression that sender name is NOT `SHA512(MAID-Public-signing-Key)` BUT `NameType(MAID-Public-signing-Key.vec())`. As I don't think it is good to have a separate ID&Pub-key exchange procedure which is outside SAFE network. Having such exchange procedure will definitely damage the user experience and hinder the spreading of this messaging tool.

proposed/MPID Messaging.md
maqi: I think it has been discussed long time ago that: any attachment in messaging, say image or video shall be stored as im_data first, then in messaging, there is only a link to data/meta_data to be used.
If there is reason that sd_data is limited as <100KB, then mpid_message shall also have the same limit, as they are in the same usage scenario.
There shall never be segmentation and re-assemble happen between SD and mpid_message.

proposed/MPID Messaging.md
ustulation: `MAID-Public-signing-Key` is [32 bytes](https://github.com/jedisct1/libsodium/blob/master/src/libsodium/include/sodium/crypto_sign_ed25519.h#L22) .. And i implemented [this](https://github.com/maidsafe/safe_client/blob/master/src/client/mod.rs#L106) after discussion with @benjaminbollen , who i think is better equipped than I to comment on this, since i don't know this 100% myself.

proposed/MPID Messaging.md
Fraser999: 1. I guess I'd seen it more like a mapping of <human user's chosen handle for contact, MPID public keys> where the handle is whatever that human wants to call the other guy in his address book, and the MPID is both the signing and encrypting public keys.
2. Yes.  The header isn't encrypted, but the `subject` can contain encrypted data if required.
3. Yes.
4. Yes.  In the future, we may have the inbox configurable by the user to say whether to allow incoming messages from any user, or add a whitelist and/or blacklist of MPID IDs.

proposed/MPID Messaging.md
Fraser999: > I think it has been discussed long time ago that: any attachment in messaging, say image or video shall be stored as im_data first, then in messaging, there is only a link to data/meta_data to be used.

I agree, but we don't (can't?) do anything to enforce this beyond that it makes sense to do this so the sender doesn't fill his outbox.

> If there is reason that sd_data is limited as <100KB, then mpid_message shall also have the same limit, as they are in the same usage scenario.

I think the main reason is that SDs are held in memory and passed around in account transfers unlike the Immutable chunks which are managed differently.

Arguably the usage scenarios are different since SD is mutable, and Messages and Headers aren't.  They *could* be handled as ImmutableData except we can't delete Immutable chunks and these don't allow different tags (we couldn't tell a Header from a Message unless we parsed the Immutable Data first).

For now, if we do wrap them in SD, then I agree that we can arbitrarily set an upper limit on Message size (Header is already limited) so it fits in a single SD.

proposed/MPID Messaging.md
Fraser999: Well, I think the MPID needs to be a standard MaidSafe key packet so it can be validated and revoked in the normal way.

For a long time, the name of the MPID packet was a human-chosen string, so the Hash of the contents *didn't* yield the name.  However, that changed (maybe a year or more ago) to the current setup.

The only way I see this damaging UX is that the contact will be a non-memorable string.  It's like Bitcoin wallet addresses, but worse because the length won't be just 35 chars, it'll be up to 128 chars depending on how we encode the binary MPID ID.

I have made a couple of suggestions in the past as to how to improve this situation, and I'm sure there are a few other ideas kicking about, so for now I don't see the exchange of MPID IDs as a big problem.

maqi: Once you receive such MPID ID and Key, it needs to be `integrated` into the MpidMessaging process, i.e. through some configure file or IPC? that will be problematic IMO.
The straight forward way is to make MPID key and NameType convertible, just expand pub key's 32B to Nametype's 64B by duplicating or filling with all zeros.

proposed/MPID Messaging.md
```diff
@@ -0,0 +1,295 @@
...
+
+## Outbox
+
+This is a simple data structure for now and will be a hash map of serialised and encrypted `MpidMessage`s.  There will be one such map per MPID (owner), held on the MpidManagers, and synchronised by them at churn events.
+
+## Inbox
+
+This is an even simpler structure and again there will be one per MPID (owner), held on the MpidManagers, and synchronised by them at churn events.
+
+This can be implemented as a `Vec<(sender: ::routing::NameType, signed_header: Vec<u8>)>`.
```
maqi: may need to change to : `Vec<(sender: Mpid, signed_header: Vec<u8>)>` ?

proposed/MPID Messaging.md
Fraser999: OK - after discussion with Qi, we think the easiest way to handle this initially is to change the type of `MessageHeader::sender` from `::routing::NameType` to `::sodiumoxide::crypto::sign::PublicKey`.

This will do away with the need for Clients and Managers to hold a copy of MPIDs.

canndrew:
Hi guys. Sorry for my noobish questions, I'm still trying to get my head around how all this works.

Under this proposal, what determines the size of an outbox? Is it the total storage that MpidClient(A) has available or is MpidManagers(A) able to set a fixed limit on the number of messages? If it's the former then it doesn't seem like it provides much protection against spam. Emails can be tiny, in the order of a kilobyte, so someone with a terabyte of storage could have a billion emails sitting in their outbox at once. There would still be a marginal cost associated with every spam sent, but it would be proportional to the cost of the storage space they require (likely very cheap) rather than to the amount of nuisance they cause. People have been talking for decades about having a system where to send an email, the sender must pay some tiny amount (determined by the receiver) to the receiver of an email. With SafeCoin we could finally implement something like that. It seems to solve the spam problem better because it means I could charge strangers some significant amount, say a dollar per message, to contact me while I could issue certificates to my friends and other contacts allowing them to contact me for free or for cheap.

The fact that the sender can see when a message has been read also raises some privacy concerns. For example, a spammer could tell who is reading their spam and what messages they're reading and use this to better target people.

Also, is this envisioned as a direct replacement for email/IM or just as a protocol that someone could build an IM client on top of? I get the impression that it's the former but if third party apps aren't able to implement this themselves then they may want to utilise it for their own messaging needs. This would mean my inbox could end up containing messages that aren't meant to be read by me but by apps that I have installed and I'd need a way to filter them.

canndrew:
Also

> If people are not picking up messages, it is because they do not want them.

this isn't necessarily true. If I go camping for a week and don't check my messages I wouldn't want to get unsubscribed from a bunch of mailing lists and not even realise that it's happened.

Fraser999:
> what determines the size of an outbox?

At the moment it's a fixed size for every MPID of 128MB, it's not configurable or adjustable.

> With SafeCoin we could finally implement something like that.

I think there's a lot of merit in that.  This proposal predates safecoin (the RFC was initially copied from old C++ documentation) hence safecoin isn't mentioned.

At the end of the day, we probably still need a network outbox and inbox so that clients can communicate without requiring both to be connected simultaneously.  And if we need such functionality, we probably also need to limit the size of each.  I see your safecoin proposal as an additional RFC which complements this one rather than a replacement (although it would change some of the rationale statements in this one).

> The fact that the sender can see when a message has been read also raises some privacy concerns.

I think future iterations will see the ability for users to whitelist or blacklist other users, hence spammers would be unable to continually target a given user if the user chose to not be targeted.  I guess another option might be to allow the receiver to "put back" the message into the sender's outbox if they mark it as spam.  So while the message would be absent for a (likely short) period of time from the spammer's outbox, it could still be used to make things difficult for them.

> Also, is this envisioned as a direct replacement for email/IM or just as a protocol that someone could build an IM client on top of? I get the impression that it's the former

At the moment it's actually the latter.  This RFC doesn't cover the long-term storage of received messages for example, so an email app would need to store received messages back to the network in some way after retrieving them if it wanted to allow the user's messages to be synced across devices.  In fact, IM clients might be better suited to creating direct Client-to-Client connections and archiving the received messages in a network shared folder.

> This would mean my inbox could end up containing messages that aren't meant to be read by me but by apps that I have installed and I'd need a way to filter them.

Yes - the intent is that the message contents and the Header subject can be a serialised anything, so different apps will encode different structures into the data field.  Your app should therefore be able to tell which messages are for it by parsing the Header.

Additionally, apps could perhaps create their own MPID on behalf of the user to ensure that the messages for the MPID are all for it.  This is probably another area where your safecoin proposal could help.  If a MPID *has* to be associated with a safecoin wallet (or given safecoin to use), there's no incentive for a malicious user/app to try and create large numbers of MPIDs to circumvent the outbox limitation.

> If I go camping for a week and don't check my messages I wouldn't want to get unsubscribed from a bunch of mailing lists and not even realise that it's happened.

Yeah - I agree, but I think the onus is on the senders to get this right.  As mentioned above, we need a network outbox, and until (or even if) we implement something more like your safecoin proposal, we need to limit the size of that outbox so the network can manage it.

proposed/MPID Messaging.md
ustulation: > easiest way to handle this initially is to change the type of MessageHeader::sender from ::routing::NameType to ::sodiumoxide::crypto::sign::PublicKey

Yes, I too think that would be better.

> This will do away with the need for Clients and Managers to hold a copy of MPIDs

This however I am unable to understand. Let me put it in detail so that you can explain better.

If the Clients have no idea of the sender's MPID's then what is the use of having a signed message ? One answer by @maqi was:
> the signed stuff here is to ensure a security between a messaging peer (on vault level).
messages will be stored in OutBox and Inbox, once it is restored and forwarded again, the content is open to be modified and routing sign can not eliminate that risk

if a message can be modified that way then one can well argue that entire message along with the Key replaced such that the whole combination `Vec<(sender: Mpid, signed_header: Vec<u8>)>` is valid. I thought in this case the Client would check `sender` against the copy of `MPID`s that he has and if it matches, he can be certain the message has not been tampered with and if it does not then he it depends if he trusts the unknown `sender`. If Clients are assumed to *not* hold MPID copies then what purpose does signing solve ?

canndrew:
> At the moment it's a fixed size for every MPID of 128MB, it's not configurable or adjustable.

Different people are going to have legitimate reasons for needing different amounts of space. Someone who runs a popular mailing list sending large messages might find that this limit makes MPID messaging unusable whereas an average user may never fill 128MB. Also, what's stopping people from creating multiple MPID outboxes to overcome the limit?

ustulation:
> Different people are going to have legitimate reasons for needing different amounts of space. Someone who runs a popular mailing list sending large messages might find that this limit makes MPID messaging unusable

From my prior discussion above, i get that this is not meant to be a permanent storage for messages. Only until the messages are picked up by Client `B` at which point the `B` sends explicit messages to clear it from his *and* `A`'s `MpidManagers`. Of-course there can be questions like what if `B` decides never to clear. Since Inbox and Outbox sizes are same and Inbox stores only the Message Headers, it may be fine for `B` to not initiate purge but the effect might magnify at `A`'s end in his Outbox (which stores message bodies too). The other concern about privacy that you mentioned is also a valid point for me. But given scenarios where things are ideal and `B` does initiate a purge of all read messages, it is unlikely that text only content might jump to as high as 128 MiB. Heavy stuffs like images etc. are just links to immutable data within the textual content from what I got during earlier discussions here.

StructureData limit is 100 KiB anyway. So maybe if the size of a message is going to breach this limit then the actual message will be some Immutable Data and this will only contain a link ? If that is the case then we can assume all (Optimally managed) StructuredData to be less than 1 KiB. That would give approximately 128000 unread/un-purged messages before you run out of space.

>  Also, what's stopping people from creating multiple MPID outboxes to overcome the limit?

My guess is MPID Outboxes are not created by the client. It is done by the vaults and an Outbox is associated with MPID keys which are unique and just one pair (Asymmetric right now) for every user. So just like one user cannot create multiple accounts using same set of credentials (hence MAID keys), a user cannot have multiple sets of MpidManagers and thus Outboxes if vaults don't allow.

canndrew:
> So just like one user cannot create multiple accounts using same set of credentials (hence MAID keys), a user cannot have multiple sets of MpidManagers and thus Outboxes if vaults don't allow.

Couldn't a spammer still create multiple identities and use it to bypass the limit?

ustulation:
He has create multiple accounts for that. Then for the network each account is a separate spammer even if it's being operated by the same person. It's probably like creating millions og gmail accounts ?

maqi:
> Of-course there can be questions like what if `B` decides never to clear. Since Inbox and Outbox sizes are same and Inbox stores only the Message Headers, it may be fine for `B` to not initiate purge but the effect might magnify at `A`'s end in his Outbox (which stores message bodies too). 

Sender `A` can send a `delete` request to cleanup its messages in Outbox. I think it's being mentioned in the section Future Work

proposed/MPID Messaging.md
maqi: maybe we can just have both: index/parent_index and timestamp  :)

proposed/MPID Messaging.md
dirvine: >I think the main reason is that SDs are held in memory
These should not be in memory though, but stored on disk in a btree like container. Perhaps this needs added to list of vault tasks?

proposed/MPID Messaging.md
dirvine: Yes we don't need such limits any more I feel.

proposed/MPID Messaging.md
```diff
@@ -0,0 +1,266 @@
...
+    ::routing::structured_data::MAX_STRUCTURED_DATA_SIZE_IN_BYTES - 512 - MAX_HEADER_METADATA_SIZE;
+pub const MAX_INBOX_SIZE: usize = 1 << 27;  // bytes, i.e. 128 MiB
+pub const MAX_OUTBOX_SIZE: usize = 1 << 27;  // bytes, i.e. 128 MiB
+```
+
+### `MpidHeader`
+
+```rust
+pub struct MpidHeader {
+    pub sender: ::sodiumoxide::crypto::sign::PublicKey,
```
maqi: In MpidManager's Inbox, MpidHeader will be stored as signed_header, and once client received signed_header the publick_key is required first to verify/un-patch the signed_header to MpidHeader format.
so the sender's public key shall not be defined in MpidHeader, but be stored in MpidMessage and be transfered among MpidManagers and to recipient via StructuredData's owner key (get_owner_keys() function)

proposed/MPID Messaging.md
```diff
@@ -0,0 +1,272 @@
...
+
+The `sender` field is hopefully self-explanatory.  The `guid` allows the message to be message to be uniquely identified, both by receivers and by the manager Vaults which need to hold the messages in a map-like structure.
+
+The `metadata` field allows passing arbitrary user/app data.  It must not exceed `MAX_HEADER_METADATA_SIZE` bytes.
+
+### `MpidMessage`
+
+```rust
+pub struct RecipientAndBody {
+    pub recipient: ::routing::Authority::Client ,
```
maqi: Authority::Client is the proxy for the vault the recipient client got connected to.
so some kind of registration (register the Authority::Client with recipient's name ) is required in Inbox MpidManagers.

proposed/MPID Messaging.md
```diff
@@ -0,0 +1,451 @@
...
+    data: XXX,
+    previous_owner_keys: vec![],
+    version: 0,
+    current_owner_keys: vec![sender_public_key],
+    previous_owner_signatures: vec![],
+}
+
+```
+
+where XXX is indicated in the following list of all MPID-related messages:
```
maqi: There is no need to wrapper actions such as `PUT`, `POST`, `DELETE` into the `data`, as when the request needs to be composed by client or vault, such action already needs to be specified. If the actions being wrapped, people will ask which `request` should be used to send out the ? always to be a PUT request, or a POST request?

proposed/MPID Messaging.md
```diff
@@ -0,0 +1,451 @@
...
+| Post[GetAllHeadersResponse[Vec<Header>]] | Sender's Managers (outbox)   ==> Sender Client |
+
+
+The various different types for `StructuredData::data` can be enumerated as:
+
+```rust
+#[allow(variant_size_differences)]
+enum MpidMessageWrapper {
+    /// Notification that the MPID Client has just connected to the network
+    Online,
```
maqi: I don't think a separate Online message needs to be presented.
A `GetAllHeaders` shall do the job and it does make more sense: when a client is interested in what's available in his Inbox, then there is meaning to push to it.

proposed/MPID Messaging.md
```diff
@@ -0,0 +1,451 @@
...
+    /// Try to retrieve the message corresponding to the included header
+    GetMessage(MpidHeader),
```
maqi: MpidHeader shall be/only needs to be the name of the message?

proposed/MPID Messaging.md
maqi: I think `data` shall always be a serialised `MpidMessageWrapper` ?

Fraser999: I don't think it has to be.  I was thinking it could be a serialised `MpidMessage` or `MpidHeader`.

proposed/MPID Messaging.md
Fraser999: Well, the original intent of the `GetAllHeaders` message was to get all headers from your _outbox_ - not inbox.  This is useful since so a client can check to see if any need deleting from its outbox.  There's not the same need for a client to be able to get all its inbox headers I don't think - except just the once at the start of a new session.

I guess we need to decide if we want to allow clients to get all inbox and outbox headers whenever they want.  If so, do we just return both the inbox and outbox headers in a single message?

Also, I'd expected the `Online` message to disappear once Routing implements an event to notify of a client coming online.

proposed/MPID Messaging.md
Fraser999: Yes - I think you're right.

proposed/MPID Messaging.md
```diff
@@ -0,0 +1,451 @@
...
+    current_owner_keys: vec![sender_public_key],
+    previous_owner_signatures: vec![]
+}
+```
+
+When handling churn, MpidManagers will synchronise their outboxes by sending each included MpidMessage as a single network message.  However, since MpidHeaders are significantly smaller, inboxes will be sent as a message containing a vector of all contained headers.
+
+## Message Flow
+
+![Flowchart showing MPID Message flow through the SAFE Network](MPID%20Message%20Flow.png)
```
maqi: The flowchart doesn't show the recipient just asking for a particular full message from sender's MpidManagers ?
Or recipient is only allowed to use GetAll ?

proposed/MPID Messaging.md
maqi: If the data field can be a serialised MpidMessageWrapper, MpidMessage, or MpidHeader, then they can not be combined into one request type, otherwise the parser will have issue unless we capture failed parse attempt and try different type one by one.
i.e. if say in a post request, the sd.data maybe a MpidMessage or a MpidHeader, then we have to do something like `if parse<MpidHeader>(sd.data) {} else { parse<MpidMessage>(sd.data); }`.

proposed/MPID Messaging.md
Fraser999: I was seeing this as the Client sends `Post(Online)` message and then gets each message as and when it can be retrieved by its managers from the network.  If the client is already connected when a new message is created for it, it will just receive the message.  So I don't think we need an explicit GetMessage or GetAllMessages for Client to own inbox.

proposed/MPID Messaging.md
Fraser999: I'd agree except that they can be told apart by looking at the sender and receiver.  So Put(Message) is always from Client, Put(Header) is always from Vault, etc.

I'm not totally against wrapping *all* MPID-messaging types just to be consistent and clear if you think it's worth it.  I guess there won't be much extra code to always wrap/unwrap `MpidMessage`s and `MpidHeader`s.

proposed/MPID Messaging.md
maqi: Ok, that means recipient will never receive mpid_header separately (i.e.no light traffic preview allowed? ) but the full message directly.

Fraser999: Yes - that was decided in Friday's meeting I believe.

madscramble:
> Someone who runs a popular mailing list sending large messages might find that this limit makes MPID messaging unusable

I think when apps developers are faced with the economics of messaging, they will come up with better schemes.  For instance, instead of such a mailing list sending a notice to every subscriber for every update, it might send one to each that indicates "update available" then stop until the subscriber requests updates, at which time he gets all the updates since his last request.  Offline subscribers aren't causing a logjam and online subscribers are cycling through this exchange quickly.  That's more tracking work for the apps (and maybe an opportunity for a service-type app to manage this), but it's nice and clean in the case of solicited notifications while still discouraging unsolicited messages.

PR #44, opened 2015-09-24
WIP: Connection management proposal.

brian-js:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/44)

### DIFF
--- a/proposed/0008-Connection_management.md
+++ b/proposed/0008-Connection_management.md
@@ -17,10 +17,10 @@ Both outgoing and incoming connections are created to peers with unknown directi
 
 1. Introduce a `State` object to `RoutingCore` representing the distinct phases of execution for network entities characterised as follows.
 
-a. Initially client/node in disconnected state, and when all connections are dropped/lost.
-b. Bootstrapping, initiated in constructor by calling crust service function and halted by crust event sent over channel.
-c. If non-client node, connected phase adds connections to routing table.
-d. In order to prevent any further network activity a terminated state.
+- Initially client/node in disconnected state, and when all connections are dropped/lost.
+- Bootstrapping, initiated in constructor by calling crust service function and halted by crust event sent over channel.
+- If non-client node, connected phase adds connections to routing table.
+- In order to prevent any further network activity a terminated state.
 
 ```rust
 pub enum Phase {
@@ -37,9 +37,9 @@ pub struct State {
 }
 ```
 
-2. Merge the `RoutingNode` functions `handle_new_connection` and `handle_new_bootstrap_connection`.
+1. Merge the `RoutingNode` functions `handle_new_connection` and `handle_new_bootstrap_connection`.
 
-3. For connections, add to utils folder a timed `ConnectionFilter` object for key type `crust::Connection`, and value, new type, `ExpectedConnection`. An object of type `ConnectionFilter<Connection, ExpectedConnection>` replaces the current `connection_filter` in `RoutingNode`.
+1. For connections, add to utils folder a timed `ConnectionFilter` object for key type `crust::Connection`, and value, new type, `ExpectedConnection`. An object of type `ConnectionFilter<Connection, ExpectedConnection>` replaces the current `connection_filter` in `RoutingNode`.
 
 ```rust
 pub struct Connection {
@@ -60,11 +60,11 @@ pub struct ConnectionFilter<K, V> {
 
 For incoming connect requests, we want to handle, store the `ExpectedConnection::ConnectRequest(ConnectRequest)` in the timed filter and try to connect. For incoming connect responses check the returned `ConnectRequest` was sent by us and store the `ExpectedConnection::ConnectResponse(ConnectResponse)` in the timed filter and try to connect. On receipt of a crust OnConnect/OnAccept event within the time limit for the stored expected `ConnectRequest/ConnectResponse` add the connection to the routing table and remove from filter.
 
-4. Update Hello.
+1. Update Hello.
 
-5. Remove `Unidentified` connections from `ConnectionName`.
+1. Remove `Unidentified` connections from `ConnectionName`.
 
-6. In the event of disconnect implement re-bootstrapping.
+1. In the event of disconnect implement re-bootstrapping.
 
 # Drawbacks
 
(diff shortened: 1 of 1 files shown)


proposed/0008-Connection_management.md
```diff
@@ -0,0 +1,82 @@
...
+1. In order to prevent any further network activity a terminated state.
+
+```rust
+pub enum Phase {
+    Disconnected(bool),
+    Bootstrapping(bool),
+    Connected(bool),
+    Terminated(bool),
+}
+```
```
Fraser999: Doesn't `Bootstrapping` imply `Disconnected`?  If so, can we remove `Bootstrapping`?

Apart from that, these states are mutually exclusive aren't they?  If so, why do we need the `(bool)` aspect to each - wouldn't a C-style enum do?

proposed/0008-Connection_management.md
```diff
@@ -0,0 +1,82 @@
...
+
+```rust
+pub struct State {
+    phase: Phase
+}
+```
```
Fraser999: It's not clear why we need to wrap the enum in a struct.  Couldn't we just rename the enum to `State` and hold an instance of that in `RoutingCore`?

proposed/0008-Connection_management.md
```diff
@@ -0,0 +1,82 @@
...
+Both outgoing and incoming connections are created to peers with unknown direction. Use the connect/accept connection code available from crust to store timed state in-line with crust connection handling changes for connections in either direction. This will allow us to determine unambiguously who the connection initiator is and act accordingly.
+
+# Detailed design
+
+## State
+
+Introduce a `State` object to `RoutingCore` representing the distinct phases of execution for network entities characterised as follows.
+
+1. Initially client/node in disconnected state, and when all connections are dropped/lost.
+1. Bootstrapping, initiated in constructor by calling crust service function and halted by crust event sent over channel.
```
Fraser999: Bootstrapping happens whenever the node is disconnected doesn't it, not just in the constructor?

proposed/0008-Connection_management.md
```diff
@@ -0,0 +1,82 @@
...
+- RFC PR: (leave this empty)
+- Issue Number: (leave this empty)
+
+# Summary
+
+Introduce the notion of clearly defined connection phases and establish directed routing network connections in conjunction with crust.
+
+# Motivation
+
+Both outgoing and incoming connections are created to peers with unknown direction. Use the connect/accept connection code available from crust to store timed state in-line with crust connection handling changes for connections in either direction. This will allow us to determine unambiguously who the connection initiator is and act accordingly.
```
Fraser999: I don't think this really explains why we need/want this functionality.  There's no reason given why keeping the state and knowing direction of connections is a good thing or solves an existing issue.

proposed/0008-Connection_management.md
```diff
@@ -0,0 +1,82 @@
...
+    peer_addr: SocketAddrW,
+    local_addr: SocketAddrW,
+}
+
+enum ExpectedConnection {
+    InternalRequest::Connect(ConnectRequest),
+    InternalResponse::Connect(ConnectResponse, SignedToken)
+}
+
+pub struct ConnectionFilter<K, V> {
```
Fraser999: Is there a reason for this to be a generic object?

proposed/0008-Connection_management.md
```diff
@@ -0,0 +1,82 @@
...
+
+```rust
+pub struct State {
+    phase: Phase
+}
+```
+
+## Connection Filter
+
+For connections, add to utils folder a timed `ConnectionFilter` object for key type `crust::Connection`, and value, new type, `ExpectedConnection`. An object of type `ConnectionFilter<Connection, ExpectedConnection>` replaces the current `connection_filter` in `RoutingNode`.
```
Fraser999: I don't think there _is_ a utils folder in Routing.

Fraser999: I think we should briefly explain what we mean by a timed filter, either here where we introduce it, or below where we describe its usage.

proposed/0008-Connection_management.md
```diff
@@ -0,0 +1,82 @@
...
+}
+```
+
+```rust
+pub struct State {
+    phase: Phase
+}
+```
+
+## Connection Filter
```
Fraser999: I can't say I like the name "Connection Filter".  I think something like "Expected Connections" better represents it.

proposed/0008-Connection_management.md
```diff
@@ -0,0 +1,82 @@
...
+## Updates to Existing Code
+
+1. Merge the `RoutingNode` functions `handle_new_connection` and `handle_new_bootstrap_connection`.
+1. Remove `Unidentified` connections from `ConnectionName`.
+1. In the event of disconnect implement re-bootstrapping.
+1. Update Hello in line with crust events.
+
+# Drawbacks
+
+N/A
```
Fraser999: We've tried this before in the C++ codebase, and I recall problems when two peers connect simultaneously (both are initiators).  We should probably cover that situation in this document.

proposed/0008-Connection_management.md
brian-js: Yes. After committing it I noticed that.

proposed/0008-Connection_management.md
brian-js: There will be a utils folder when the implementation gets done. I'll make that clearer.

proposed/0008-Connection_management.md
brian-js: The idea behind the ExpectedConnections, ConnectionFilter above, is that we will know. I'll try to make the intention clearer before committing without the WIP.

proposed/0008-Connection_management.md
brian-js: Maybe this needs some clarification on my part. My understanding is that initially we're disconnected. Bootstrapping is initiated in the constructor and stops when crust signals it up or routing calls stop. If we're a node we should enter a connected phase. If we lose all connections we're disconnected and should try bootstrapping again.

proposed/0008-Connection_management.md
brian-js: Agreed on the explanation.

PR #46, opened 2015-09-28
Adds UDP hole punching procedure proposal

inetic:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/46)

### DIFF
--- a/proposed/0000-UDP-hole-punching.md
+++ b/proposed/0000-UDP-hole-punching.md
@@ -67,7 +67,7 @@ And the result of such call shall be an event `OnUdpSocketMapped` holding a stru
     struct MappedUdpSocket {
         result_token: u32,
         udp_socket: UdpSocket,
-        public_address: SocketAddr, // of node A
+        public_addresses: BTreeSet<SocketAddr>, // of node A
     }
 
 Once upper layers receive such event, they can send/route `MappedUdpSocket::public_address`
@@ -243,10 +243,18 @@ impl PeriodicSender {
 impl Drop for PeriodicSender;
 
 fn blocking_get_mapped_udp_socket(request_id: u32, helper_nodes: Vec<SocketAddr>)
-    -> (UdpSocket, Result<SocketAddr>)
+    -> (UdpSocket, BTreeSet<SocketAddr>)
   const TIMES_TO_SEND = 5;
   let udp_socket = UdpSocket::bind("0.0.0.0:0");
+  let mut addrs = BTreeSet::new();
   let periodic_sender = PeriodicSender::new(udp_socket);
+
+  // This happens when `sort_helping_nodes_by_preference` returns an empty
+  // vector. An empty vector is given when only nodes from the same network are
+  // found.
+  if helper_nodes.len() == 0 {
+    // use `getifaddrs` to return the addresses of our interfaces
+  }
 
   for cx in helper_nodes {
     periodic_sender.start(cx, TIMES_TO_SEND, GetExtAddr::new(request_id));
@@ -262,10 +270,11 @@ fn blocking_get_mapped_udp_socket(request_id: u32, helper_nodes: Vec<SocketAddr>
           datagram.ext_address
         }
       }
-      return (udp_socket, Ok(our_ext_address));
+      addrs.insert(our_ext_address);
+      return (udp_socket, addrs);
     }
   }
-  (udp_socket, Err(...))
+  (udp_socket, addrs)
 }
 
 // The non blocking version that users of the Crust library will use.
@@ -276,8 +285,8 @@ pub fn Service::get_mapped_udp_socket(&self, result_token: u32) {
     let helpers = self.sort_helping_nodes_by_preference();
 
     thread::spawn(move || {
-      let (socket, mapped_address) = blocking_get_mapped_udp_socket(request_id, helpers);
-      event_sender.send(Event::OnUdpSocketMapped(MappedUdpSocket::new(result_token, socket, mapped_address)));
+      let (socket, mapped_addresses) = blocking_get_mapped_udp_socket(request_id, helpers);
+      event_sender.send(Event::OnUdpSocketMapped(MappedUdpSocket::new(result_token, socket, mapped_addresses)));
     });
   });
 }
@@ -330,17 +339,19 @@ pub fn Service::udp_punch_hole(&self,
                                result_token: u32,
                                udp_socket: UdpSocket,
                                secret: Option<[u8,4]>,
-                               peer_addr: mut SocketAddr /* of  node B */) {
+                               peer_addrs: mut BTreeSet<SocketAddr> /* of  node B */) {
   Self::push(&self.cmd_sender, move |state| {
     let event_sender = state.event_sender.clone();
 
     thread::spawn(move || {
-      let (socket, peer_addr) = blocking_udp_punch_hole(udp_socket,
-                                                        secret,
-                                                        peer_addr);
-      event_sender.send(Event::OnHolePunched(HolePunchResult::new(result_token,
-                                                                  udp_socke,
-                                                                  peer_addr)));
+      for peer_addr in peer_addrs {
+        let (socket, peer_addr) = blocking_udp_punch_hole(udp_socket,
+                                                          secret,
+                                                          peer_addr);
+        event_sender.send(Event::OnHolePunched(HolePunchResult::new(result_token,
+                                                                    udp_socke,
+                                                                    peer_addr)));
+      }
     });
   });
 }


proposed/0000-UDP-hole-punching.md
```diff
@@ -0,0 +1,192 @@
...
+layers are ready for the actual hole punching.
+
+## Hole punching
+
+The act of hole punching shall be initiated by a function with the following signature:
+
+    Service::udp_punch_hole(udp_socket : UdpSocket,
+                            secret: Option<[u8; 4]>,
+                            peer_addr : mut SocketAddr /* of node B */,
+                            callback: FnBox<(UdpSocket, io::Result<SocketAddr>)>)
```
vinipsmaker: `Service::udp_punch_hole` doesn't know enough info to initiate a connection. Currently, only one UDP-based protocol is supported, then it's not a problem. But given this function will be used by upper layers, we should do a little more to avoid breaking API as soon as we implement rUDP or whatever other custom protocol builtin on top of UDP.

What about adding another param like `Protocol`? Given not all `Protocol`s are UDP-based, we could have a new enum, `UdpProtocol`.

proposed/0000-UDP-hole-punching.md
```diff
@@ -0,0 +1,192 @@
...
+`X` connects to `C(X)`.
+
+In the following text we’ll describe steps taken at node `A`, but the same steps need to
+be taken at node `B`.
+
+For `A` to find out its external address `A` first needs to create a UDP socket `S(A)`
+with an arbitrary local port.  `A` will then use `S(A)` to periodically send
+datagrams to `U(C(A))` containing a MAGIC number and the ID of the request. Each time
+`C(A)` receives such datagram, it sends back a datagram containing public endpoint
+of `S(A)` together with the request ID.
```
vinipsmaker: It's not clear to me if this proposal requires crust peers to listen on yet another port. Is there any issue if we relay on the internal message infrastructure?

proposed/0000-UDP-hole-punching.md
```diff
@@ -0,0 +1,192 @@
...
+is limited in number of NAT types it can successfully punch through.
+
+# Alternatives
+
+Another approach would work with the use of multiplexing. That is, if we had a UDP socket
+which is already connected to a remote peer, we could also use it to communicate
+with other peers as the hole has already been punched. Problem with this approach is 
+that it would only help with `Full-cone NAT` types because it is the only NAT
+type where a hole punched to one peer/host can be reused with another peers/hosts.
+ resulting in slower code (re
```
vinipsmaker: > slower code (re

Looks like there is missing text here

proposed/0000-UDP-hole-punching.md
inetic: It is not necessary the case that this function will be used by the upper layers. This function is just a minimal API that is needed to punch a UDP hole in a NAT. We should then be able to easily wrap other functionality on top of this function (such as initiate uTP or UDT rendezvous connections).

inetic: > What about adding another param like Protocol? Given not all Protocols are UDP-based, we could have a new enum, UdpProtocol.

We can add other functions with these additional parameters, but having the core function `udp_punch_hole` with the minimal API that "does one thing and does it right" and make the other functions use of this one is better design IMO.

proposed/0000-UDP-hole-punching.md
inetic: Yes, we'll need to listen on yet another port. We cannot rely on existing connections, because:

1. We need to listen on UDP port, and already established connections may be using TCP
2. We're asking what is the (address, port) of the newly created socket (`S(A)`) not of the socket that is already connected to the peer.

vinipsmaker: Okay

proposed/0000-UDP-hole-punching.md
```diff
@@ -0,0 +1,192 @@
...
+   enough to express all different combinations asynchronous programming has to offer.
+   I.e. coroutines only allow for sequentinal combination of coroutines,
+   more advanced patterns (such as "start many async actions and continue when
+   all of them finish") need to be complemented with other approaches.
+
+3. Futures: these (AFAIK) have bigger combinatorial power than coroutines and
+   many libraries already ship with these patterns abstracted (think of functions
+   like `when_any` or `when_all`). That said, they have the disadvantage
+   that they combine blocking and non blocking paradigms, resulting in 
+   slower code (read [this](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3747.pdf) for
```
vinipsmaker: replace n3747 by newer [n4045](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4045.pdf)

proposed/0000-UDP-hole-punching.md
```diff
@@ -0,0 +1,192 @@
...
+## Finding an external mapping of a UDP socket
+
+Suppose, we have two nodes `A` and `B`, where each one is possibly behind a NAT and
+they want to connect to each other.  Routing needs to send `A` an endpoint of `B` and
+vice versa. But `A` and `B` don’t know their public
+endpoints a priori (they may know their IP addresses from previous connection establishments,
+but not their ports). Each node therefore needs to create a UDP socket and use it to ask some
+other node it is connected to what the public endpoints of the UDP socket is. We’ll refer
+to such helping nodes `C(A)` and `C(B)` (it needs not be the same node for `A` and `B`).
+We'll call `U(C(X))` the UDP port on which X can contact `C(X)` in order to find its public
```
vinipsmaker: Should "_the UDP port_" be "_the UDP address_"?

proposed/0000-UDP-hole-punching.md
```diff
@@ -0,0 +1,192 @@
...
+endpoint. `X` finds out about `U(C(X))` through the initial handhsake, that is, when
+`X` connects to `C(X)`.
```
vinipsmaker: > `X` finds out about `U(C(X))` through the initial handhsake, that is, when
`X` connects to `C(X)`

As there are some nodes which cannot act as helping nodes, shouldn't this handshake data be optional? Like the following.

```rust
struct Handshake {
    ucx: Option<SocketAddr> // `U(C(X))`
}
```

proposed/0000-UDP-hole-punching.md
inetic: I used "UDP port" because we already know the address of `C(X)` (if we didn't know it, we wouldn't have been able to connect to it in the first place).

proposed/0000-UDP-hole-punching.md
vinipsmaker: Resolved in https://github.com/inetic/rfcs/commit/7db7a35d06cea9bbe6875a4bd844c05a57f78a70

proposed/0000-UDP-hole-punching.md
vinipsmaker: Okay

proposed/0000-UDP-hole-punching.md
inetic: The problem is that `C(X)` doesn't know whether it can be used as a helping node. To know that, ti would also need to know:

1. Is it on LAN
2. Is it behind a NAT, if so, what kind of NAT (would need to either support UPnP or be a Full-cone-NAT)
...

I'm currently thinking about the exact requirements when `C(X)` can be a helper of `X`.

vinipsmaker: > The problem is that `C(X)` doesn't know whether it can be used as a helping node.

I thought about that too, but wanted to have the possibility anyway. But you're right, its a logic problem.

vinipsmaker: What about the following structure then?

```rust
struct Handshake {
    ucx: Vec<SocketAddr> // `U(C(X))`
}
```

The node will try to add port mapping to this socket through UPnP IGD, but even if the IGD device returns success, there isn't a guarantee that the operation worked (and worse: another application from another computer is listening on the returned socket address and we probably need to add magic random numbers to add some reliability).

But maybe the node isn't behind a NAT even when an IGD device is available. In this case, the IGD device shouldn't be used. It's difficult to decide whether we should return the public mapped endpoint from the IGD device or our public address detected from the P2P network + port reported by the OS.

This structure solves the problem by allowing us to report all "public endpoints to the same socket". Then the other node can try to reach use through several socket addresses.

inetic: It should contain only the ports without IP addresses
```rust
struct Handshake {
    ucx: Vec<u16> // `U(C(X))`
}
```
We can easily find out the IP with this https://doc.rust-lang.org/nightly/std/net/struct.TcpStream.html#method.peer_addr

vinipsmaker: But there might be more than one network card on the hardware running crust. What if IGD is found on one of the network devices, but `X` is connected to `C(X)` using the other network card (which may have a different public IP)?

inetic: You got me thinking there, but I still don't think it is needed. As I see it we have two scenarios: Fist one is when `C(X)` is not behind a NAT, this is simple and `X` only needs to consult `peer_addr` to get `C`'s address. The second one is if `C(X)` is behind a NAT that is also an IGD device, in such case `X` is interested in knowing the address of the NAT, not the actual interface address of `C(X)`.

vinipsmaker: If the scenario where no more than one network card is possible, then the use of socket address instead socket port is not more robust in any way.

But shouldn't we allow this extra bit of robustness, even if very few users will experience it? Also, the added complexity is very little for this change.

inetic: If there is a way to make it more robust, then yes. It might be that I just don't see/understand it. How I understand what you're saying is that *if* `C(X)` is behind a NAT with IGD, then we should send `X` the local IP of the interface which `C(X)` uses to communicate with the NAT. This seems pointless, because `X` will never be able to connect to `C(X)` using its local IP.

vinipsmaker: It'd contain the public IP discovered. For the mapped IGD address, the IGD reported address. For the non-IGD version, it'd use the IP reported by other peers.

proposed/0000-UDP-hole-punching.md
```diff
@@ -0,0 +1,191 @@
...
+
+And the result of such a call shall be an invocation of the callback.
+
+where `MappedUdpSocket` is defined as
+
+    struct MappedUdpSocket {
+        udp_socket: UdpSocket,
+        public_address: SocketAddr, // of node A
+    }
+
```
Fraser999: The client (Routing in our own case) needs to pass a token down to Crust when making this call so that it can be passed back in the `MappedUdpSocket` I believe.

If the client were to pass down the public encryption key (`::sodiumoxide::crypto::box_::PublicKey`) of the peer it's trying to connect to via this new socket, then not only would it serve the purpose of being a token for the client, but it could also be used to encrypt *all* communications with the peer - from handshaking onwards.

If this were the case, we could also do away with the `secret` field in the hole-punching functions/structs if we wanted.

I'm not suggesting that we actually implement the encryption as part of this RFC; just that if we need to receive a token from Routing anyway, we might as well make it a public key which is unused by Crust for now other than to pass back to Routing once the connection is made.

proposed/0000-UDP-hole-punching.md
inetic: > It'd contain the public IP discovered. For the mapped IGD address, the IGD reported address.
So it would be the external address of the IGD device, no? In most cases this would be the same as what `peer_addr` will give us if we're already connected to `C(X)`. The only case where this isn't true is if `C(X)` had two interfaces, and each of these interfaces was connected to a different gateway (NAT). I think this is really uncommon for PCs. If it was just a simple addition which would make the code more robust but wouldn't make the other cases more complicated then I would like it, but with my current understanding I think it will just complicate the most common cases.

For example, if we have the non-IGD version, you suggest to use IP reported by other peers, but that sounds like a "chicken and an egg" thing, that is, for other peer to tell you your address, you must first know his address (so that you can ping him so that he can ping you back with your external address in the message).

Another problem: in case of IGD, there is another scenario where one could have one IGD device behind another NAT, in such case `peer_addr` would give better result.

And last that comes to my mind: If no IGD device has responded yet, and no one told us our external endpoint, what IP do we send back?

I think, in general, sooner or later we should implement your idea, it's just that I think we don't have all corner cases sorted yet, so I would prefer to keep things simple before we do.

proposed/0000-UDP-hole-punching.md
inetic: I should change the line "The above protocol shall be initiated from upper layers by calling", as really the plan is to use this function only internally. As it is a function taking a callback, it doesn't really need any token. However, the public API shall indeed use tokens (I just thought I would wait for specific requirements). Though my intention with this RFC is to create a *minimal* requirements for input and output, I believe that will make it easier to *reuse* this function in different contexts. So, say if we went with the idea that the PublicKey is the token, we would move the PublicKey to the callback and use it when creating the event for the user.

However I don't think using PublicKey as a token is a good idea, for these reasons:

1. It is a dependency injection: It forces the user to be aware `crypto::PublicKey`, this suits our needs, but I don't think it applies for everyone.
2. You would not be able to call the function more than once asynchronously for one remote peer.
3. It would be nice to have one token *type* for all async actions, and the type `crypto::PublicKey` would be (at best) strange for actions such as "accept connection" or "map UPnP endpoints".

Though if we have the requirement that we'd want to obfuscate this part of the protocol, then adding `crypto::PublicKey` to the function arguments would be useful, just not as a result token IMO.

proposed/0000-UDP-hole-punching.md
vinipsmaker: After all this discussion, I feel that we're "in the same line".

> but with my current understanding I think it will just complicate the most common cases

I was thinking that would be a simple addition.

> And last that comes to my mind: If no IGD device has responded yet, and no one told us our external endpoint, what IP do we send back?

It'd be useless to send the IP from the local network, but we can still send an empty vector and only use this node to connect to other nodes (instead accepting new connections). Also, maybe this is a computer with a public IP, so the IP might be sorted correctly from the start.

> I think, in general, sooner or later we should implement your idea, it's just that I think **we don't have all corner cases sorted yet**, so I would prefer to keep things simple before we do.

You're right, we need to think more about the problem before proposing a half-baked solution.

Maybe put a "future improvements" section on the RFC? So we won't forget about it?

vinipsmaker: > Maybe put a "future improvements" section on the RFC? So we won't forget about it?

I collected all the relevant notes from this discussion (and more on other communication channels) and submitted them for inclusion:

- https://github.com/inetic/rfcs/pull/2
- https://github.com/inetic/rfcs/pull/3

It should be good now.

PR #57, opened 2015-10-27
updating RFC - v1.0 Launcher JSON-API-Doc

ustulation:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/57)


dirvine:
Perhaps RSA -> Asymmetric (in our case curve 2559). Maybe we should fix the type to the 2559 type from libsodium for clarity?

ustulation:
Ah thanks @dirvine .. i had overlooked that .. corrected now hopefully

PR #63, opened 2015-11-03
Crust refactor

canndrew:
This is a work-in-progress. The meat of the RFC is yet to be written.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/63)


vinipsmaker:
- I don't think the "detailed design" is really detailed design. Maybe renaming the section to "guidelines" would do it. But probably you're going to fill more details there.
- The _motivation_ section is **very** well written, with good arguments, representative and objective examples, etc. Congratulations. Maybe a little more formatting may help even further, but there is already enough content.
- _Detailed design_ isn't too objective, in contrast to the _motivation_ section.
- Saying that threads are bad in _motivation_, but they are okay in Rust in _detailed design_ is a little contradictory. And mentioning asynchronous concepts and non-threaded implementations (e.g. `epoll`) just adds up.
- I'd suggest to either, defend synchronous or asynchronous design and, later, defend a specific model of the design (like coroutines vs promises in the case of asynchronous end up being defended).
- I'd add a "must be ready today" item into the guidelines. There isn't a "Boost.Asio for Rust" and we shouldn't wait for one.

canndrew:
> Saying that threads are bad in motivation, but they are okay in Rust in detailed design is a little contradictory. And mentioning asynchronous concepts and non-threaded implementations (e.g. epoll) just adds up.

To be clear, I don't think threads are bad, in fact I would much prefer to use real threads rather than the fake threads of something like mioco. What I think is bad is *unnecessary* threads. Threads come with a cost and using 11 different threads to read on 10 different sockets simultaneously (like the current design does) is unnecessary. You can do the same thing with 1 thread and epoll. 

By analogy, I don't think crust should strive to be a zero-allocation library like mio but we should definitely use borrowing where possible to avoid *unnecessary* allocation.

> defend a specific model of the design

I've written this: https://github.com/canndrew/crust-experiment/blob/master/src/service.rs to show the sort of thing I'm thinking of. It runs, but all it does it allow you to accept tcp connections and write to them. The important part is the style of concurrent programming that I'm aiming for with the `Acceptor`, `AcceptorReactor` and `AcceptorController` types. I'll try and write up a better explanation/motivation when I get the time.

vinipsmaker:
> By analogy, I don't think crust should strive to be a zero-allocation library like mio but we should definitely use borrowing where possible to avoid unnecessary allocation.

Yes, I don't want to the concerns over performance to hamper readability and maintainability of the project too.

> I've written this: https://github.com/canndrew/crust-experiment/blob/master/src/service.rs

Gonna read later.

proposed/0000-crust-refactor/0000-crust-refactor.md
```diff
@@ -0,0 +1,501 @@
...
+
+    let mut data = [0u8; 256];
+    let (read_go, _k) = stream.read(&mut data[..]);
+    let n = read_go.go_deadline(deadline);
+
+    println!("data == {:?}", data);
+}
+```
+
+Simple, right? For a richer example, consider how we implement this procedure
```
dirvine: I see this as where the rotor design seems to fit nicely with enum variants arranged in a state machine pattern. There it seems the time-out and any other context may be passed to child state machines. I am not saying we should do this, but it is interesting to contrast these patterns and as you have said it would be interesting to do a minimal example of each. Interesting never the less.

proposed/0000-crust-refactor/0000-crust-refactor.md
```diff
@@ -0,0 +1,501 @@
...
+                {
+                    let mut notify_write = self.acceptor.notify_write.lock().unwrap();
+                    try!(notify_write.write_all(&[::NOTIFY_WAKE])
+                                     .map_err(|e| AddListenerError::Notify {cause: e}));
+                }
+                {
+                    let mut inner = self.acceptor.inner.lock().unwrap();
+                    let token = mio::Token(inner.listeners.len());
+                    try!(inner.poll.register(&listener, token, mio::EventSet::readable(), mio::PollOpt::level())
+                                   .map_err(|e| AddListenerError::Register {cause: e}));
```
dirvine: Why not edge triggered, is there a reason to be reminded of the event ?

canndrew: Because `AcceptGo::go` returns after accepting one connection. If we receive two connections consecutively then we want to be notified about the second.

dirvine: I thought edge triggered did trigger every socket for each message received? Or do you mean the listener may get two connections per poll event? (just getting used to mio myself)

canndrew: > Or do you mean the listener may get two connections per poll event?

Right, the socket might get a second connection before accepting the first. In that case `poll` won't trigger after it re-enters.

canndrew:
I implemented most of the transport library described in the RFC and tried to write some programs using it. In practice, it's not nearly as nice to use as I'd thought it would be and so I've made major changes to the RFC w.r.t. how the APIs are designed. Specifically w.r.t. how the cancellable blocking call stuff works.

Before, all blocking calls were performed in two phases: firstly the user creates a pair of objects `(go, kill)`, then `go` can be used to perform the blocking action while the user can drop `kill` asynchronously to unblock the action. This API style works, but for anything more than trivial programs it becomes really cumbersome to manage all the `kill` objects that you end up creating.

So here's a much better design: before doing any blocking calls the user creates a pair of objects `(handle, kill)`. The `handle` object is passed as an argument to all the blocking calls. In the meantime, the user can drop the `kill` object to unblock all actions that are using the corresponding `handle`. This means there's only a single `kill` object that they need to manage which can be used to pull down a whole tree of computations. It also makes their code much nicer as they don't need to split every blocking call into two separate lines of code.

Another thing: `Acceptor` is now split into two types `StreamListener` and `ListenerSet`.  This is to make it more like `ReadStream` vs `ReaderSet` etc.  where there are different types for operating on only a single stream/endpoint and a set of them. The name has changed from `Acceptor` to `*Listener` to make it more like `TcpListener`, `UtpListener` etc.

proposed/0000-crust-refactor/crust-library.md
```diff
@@ -0,0 +1,250 @@
...
+    /// Enable to beacon.
+    fn enable(self) -> ServiceBeaconStateEnabled,
+    /// Disable to beacon.
+    fn disable(self) -> ServiceBeaconStateDisabled,
+}
+
+/// This is because some code may want a static guarantee that the beacon is held in a particular
+/// state. Usually, enforcing this sort of thing at the type-level would require typestate or
+/// dependent types. But as the state is only a boolean we can acheive this in rust by treating the
+/// states as two different types.
```
afck: To avoid duplicating the `struct ServiceBeaconState`, it might make sense to use a pattern like in [Hyper's `Response`](http://hyper.rs/hyper/hyper/server/response/struct.Response.html), with a generic parameter to distinguish between the states:
```rust
impl ServiceBeaconState<Dynamic> {
    fn is_enabled(&self) -> bool
    fn set_period(&mut self, period: Duration)
    fn enable(self) -> ServiceBeaconState<Enabled>
    fn disable(self) -> ServiceBeaconState<Disabled>
}

impl ServiceBeaconState<Enabled> {
    fn set_period(&mut self, period: Duration)
    fn disable(self) -> ServiceBeaconState<Disabled>
}

impl From<ServiceBeaconState<Enabled>> for ServiceBeaconState<Dynamic>
```

proposed/0000-crust-refactor/crust-library.md
```diff
@@ -0,0 +1,250 @@
...
+    pub fn iter_endpoint_cache<'c>(&'c self) -> EndpointCacheIterator<'c>
+}
+```
+
+The service's beacon state (enabled or disabled) is controlled via a separate
+object. This is just to take advantage of the fact that - being a simple
+boolean - it's state can be represented at the type level. The user can enforce
+compile-time guarantees about the state of the beacon during particular parts
+of their code, or they can just stick the `ServiceBeaconState` in a `Mutex` and
+use it the same way they use the `ServiceController`.
```
afck: It seems a bit inconsistent that `ServiceBeaconState` is treated like an immutable boolean value with respect to enabled/disabled, but provides a setter for the period. E. g. implementing something like `fn toggle(state: &mut ServiceBeaconState)` is impossible, whereas writing `fn adjust_period(state: &mut ServiceBeaconState)` is straightforward.

And I think putting the state in a `Mutex` is pretty cumbersome (but maybe I'm doing it wrong): To move it there you have to disassemble the `Service` first (it can't keep ownership of the object), and I think e. g. enabling the beacon is only really feasible with a `Mutex<Option<ServiceBeaconState>>` because before calling `enable`, you need to move the object out of the `Mutex`:
```rust
let mut guard = state_mutex.lock().unwrap();
if let Some(state) = mem::replace(&mut *guard, None) {
    *guard = Some(state.enable().into());
}
```

Probably the easiest solution is to just offer both:
```rust
    fn enable(&mut self);
    fn to_enabled(self) -> ServiceBeaconStateEnabled;
```

canndrew:
@afck I agree with everything you've said and I've updated the RFC.

canndrew:
I've updated the RFC to relect the fact that we're going to be using mioco.  This means removing all of our own event-loop stuff (`{Reader,Writer}Set` etc).

I've also removed all stuff to do with cancellable blocking calls (`BopHandle` and friends). Although I still think something like this would be very useful it's not clear to me how it should be implemented with mioco. It's possible that something like this could be implemented as part of mioco itself and I plan on having a chat with the mioco maintainer about it.

vinipsmaker:
> I've updated the RFC to relect the fact that we're going to be using mioco

😂

Gonna check the update later on.

canndrew:
I notice this has been merged, which is odd because we've already started refactoring crust and in some places we've taken directions that are quite different to what's outlined in this RFC. The beacon library, for instance, is now called `service_discovery` and has a different API to what's described here. Other parts would now be impossible to implement given the directions taken in the last week, such as splitting nat traversal into a separate crate. There's also been next to no feedback or discussion on this and the refactor work instead seems to be happening on a make-it-up-as-we-go-along basis.

So I don't think it makes any sense to merge this. It might make sense to merge it if it were to be heavily rewritten and then discussed and evolved into something that we can all agree on. And then only merged with the intention of implementing it.

dirvine:
It is OK as it's merged to proposed. We can accept or reject from here, but may make sense to pull this one if it could change dramatically though,.

PR #70, opened 2015-11-10
Config file for Vault

maqi:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/70)

### DIFF
--- a/proposed/vault-config-file/vault-config-file.md
+++ b/proposed/vault-config-file/vault-config-file.md
@@ -1,0 +1,87 @@
+- Feature Name: ImmutableData naming base on type
+- Type: New Feature
+- Related Components: [safe_vault](https://github.com/maidsafe/safe_vault), [launcher](https://github.com/maidsafe/safe_launcher)
+- Start Date: 10-11-2015
+- RFC PR:
+- Issue number: Proposed -
+
+# Summary
+
+This RFC outlines the config file that being required by the vault to initialise personas.
+
+# Motivation
+
+## Rationale
+
+During the disscussion [safecoin implementation](https://github.com/maidsafe/rfcs/issues/61) of [RFC SafeCoin Implementation](https://github.com/maidsafe/rfcs/blob/master/proposed/0012-safecoin-implementation/0012-safecoin-implementation.md), it is being assumed that a pmid_node persona needs to be aware of it's owner's wallet address.
+
+In addition to this, when chunk_store being intialized (currently being used by PmidNode and StructruedDataManager personas only), a configurable max usable disk space needs to be specified (currently a fixed value of 1GB is being assumed to be used as default). 
+
+To resolved the above issues, a separate config file for vault is proposed to allow user configurable parameters can be loaded and keeps retained across session (restarting of vault).
+
+## Supported Use-Cases
+
+1. User can specify the storage space for pmid_node and sd_manager
+
+1. User can specify the account to be rewarded for the service provided to SAFE network
+
+1. In case of restart, the configuration keeps the same
+
+
+## Expected Outcome
+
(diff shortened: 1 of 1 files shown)


proposed/vault-config-file/vault-config-file.md
```diff
@@ -0,0 +1,87 @@
...
+
+This RFC outlines the config file that being required by the vault to initialise personas.
+
+# Motivation
+
+## Rationale
+
+During the disscussion [safecoin implementation](https://github.com/maidsafe/rfcs/issues/61) of [RFC SafeCoin Implementation](https://github.com/maidsafe/rfcs/blob/master/proposed/0012-safecoin-implementation/0012-safecoin-implementation.md), it is being assumed that a pmid_node persona needs to be aware of it's owner's wallet address.
+
+In addition to this, when chunk_store being intialized (currently being used by PmidNode and StructruedDataManager personas only), a configurable max usable disk space needs to be specified (currently a fixed value of 1GB is being assumed to be used as default).
```
dirvine: Should this 1Gb instead be 80% of free space (continual 80% of free space) as this will allow more default installs that perhaps do not need tweaked, but provide a nice enough default performance for the network.

proposed/vault-config-file/vault-config-file.md
```diff
@@ -0,0 +1,87 @@
...
+The config file shall contains following items:
+	1. wallet_address : the associated address that shall get rewarded for the service provided, in a format of hex code
+	1. max_space_pmid_node : the max disk space allocated for pmid_node, measured in Byte
+	1. max_space_sd : the max disk space allocated for structured_data manager, measured in Byte
+
+A sample config file may looks like (1GB for pmid_node and 100MB for sd_manager):
+```rust
+{
+	wallet_address : 245df3245df3245df3245df300000000000000001cc0dd1cc0dd1cc0dd1cc0dd
+	max_space_pmid_node : 1073741824
```
dirvine: I wonder if this level of granularity is confusing for users? Maybe a single space measure is best and simpler for users. I can see why you suggest this, but perhaps if needed then it can be an advanced setting ?

proposed/vault-config-file/vault-config-file.md
maqi: I remember it has been argued that there are two issues blocking the use of percentage of free space :
1, running multiple vaults on a single machine will make them all believe having access to the full free space, which will expose a higher risk of failure.
2, I am not sure but having the impression that there is an issue in the code of calculating free space, especially to get it supported cross platform.

proposed/vault-config-file/vault-config-file.md
maqi: so just max_space being measured in MB or percentage of free space, then internally we have a fixed split ratio for immutable_data and structured_data, say 3:1 ?
I agree, this will also provide the convenience when later on we need space for other purpose say memory_map, so we only need to adjust the internal ratio, without asking the user to change the config file.

proposed/vault-config-file/vault-config-file.md
dirvine: I mostly agree, however, perhaps if each vault was continually testing free space on each store then even several vaults running with a % would be OK ? 

On code to check free space there was a few changes in rust to allow cross platform filesystem handling. It is worth checking this out again.

dirvine:
I think this is complete now @maqi  and vote for @Viv-Rajkumar  to merge and set up issue etc. Should be a nice and simple enough addition.

proposed/vault-config-file/vault-config-file.md
maqi: created a section in Alternative to note down the idea of using percentage of free space as max_space

PR #71, opened 2015-11-11
Detailing in coin and account management

maqi:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/71)

### DIFF
--- a/proposed/0012-safecoin-implementation/0012-safecoin-implementation.md
+++ b/proposed/0012-safecoin-implementation/0012-safecoin-implementation.md
@@ -145,9 +145,51 @@ This process is outlined as:
      - The safecoin close group then send a receipt message to the wallet address to inform the user
        of a new minted safecoin allocated to them.
 
+## Safecoin Management
+
+Each safecoin is represented as a piece of data held by the group closest to it's ID. Safecoin data structure is defined as:
+'''rust
+ID: 64 bytes
+OWNER: 64 bytes
+'''
+The ID of a safecoin is 64Bytes long, with the most meaningful 32 bits being sequenced index starts from 0 to 4294967295, and the left over part to be fullfiled with all zeros or other pre-defined pattern (to allow coin division).
+The OWNER of a safecoin is the wallet address provided by the pmid_node as mentioned above.
+
+The Safecoin Management group can only approve a farming request when no such targeted safecoin data has been created before.
+
+When being asked to transfer the owevership, the request transcript must provide a valid signature that can be verified by the stored OWNER, which is actually a public-key. And the the owner will be updated to the new owner.
+
+When being asked to burn a coin, the request can only be from the current owner. The piece of data will then be removed.
+
+## Account Management
+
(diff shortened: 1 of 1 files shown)


proposed/0012-safecoin-implementation/0012-safecoin-implementation.md
```diff
@@ -145,10 +145,52 @@ This process is outlined as:
...
+OWNER: 64 bytes
+'''
+The ID of a safecoin is 64Bytes long, with the most meaningful 32 bits being sequenced index starts from 0 to 4294967295, and the left over part to be fullfiled with all zeros or other pre-defined pattern (to allow coin division).
+The OWNER of a safecoin is the wallet address provided by the pmid_node as mentioned above.
+
+The Safecoin Management group can only approve a farming request when no such targeted safecoin data has been created before.
+
+When being asked to transfer the owevership, the request transcript must provide a valid signature that can be verified by the stored OWNER, which is actually a public-key. And the the owner will be updated to the new owner.
+
+When being asked to burn a coin, the request can only be from the current owner. The piece of data will then be removed.
```
dirvine: I think expend this a little. So the Burn / `Delete` is signed by the user and passed via the Client Managers who can increase the users storage space at this amount.

proposed/0012-safecoin-implementation/0012-safecoin-implementation.md
```diff
@@ -145,10 +145,52 @@ This process is outlined as:
...
+
 ## Bootstrap with clients
 
 Although there has been hostility from the community with regard to "something for nothing" approach, there is a necessity for a bootstrap mechanism. As no safecoin can be farmed until data is uploaded there is a cyclic dependency that requires a resolution. To overcome this limitation this RFC will propose that every new account created is initialised with 50 safecoins. This may be temporary and only used in test-safecoin, but it is likely essential to allow this for the time being. It may be a mechanism to kickstart the network as well.
 
+As such safecoin data needs to be presented at the moment when to be claimed as a start-up credit for a account, client's bootstrap coins are actually from a pre-injected pool : coins indexed from 0 to 67108864 are special pool that they can be claimed when free, but not allowed to be transferred, and can only be burned.
+When a new account has been created, it can take the coins having the same head-bits with that account's hash order :
```
dirvine: I think this can be added as a possibility during testing, but initially we can allow the client accounts to be as though they had used 50 safecoin and use an upgrade to implement a limited client startup.

maqi: If we say the account is initialized with safecoin, then we have to carry out similliar process to `inject` such coins from nothing.
But if we say the account having two parts: coin and storage allowance, then the account is initialized with 0 coins but certain number of storage allowance, then we don't need to do so.

dirvine: Yes I agree, there will be no safecoin here at all, but a new account created as though it had 50 safecoin, so the storage is allocated as though it was :-)

maqi:
This PR is now being replaced by the PR https://github.com/maidsafe/rfcs/pull/72

PR #75, opened 2015-11-23
Launcher API - data types (Structured Data and ImmutableData)

krishnaIndia:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/75)

### DIFF
--- a/proposed/0016-launcher-data_types-api/0016-launcher-data-types-api.md
+++ b/proposed/0016-launcher-data_types-api/0016-launcher-data-types-api.md
@@ -261,4 +261,18 @@ then the data associated with the DataMap is returned
 # Drawbacks
 If a big file/data has to be saved from the application (Say 5gb). Then
 writing the entire content of the data through the API at one go would choke up the resources.
-As a workaround the `update raw data` API can be invoked in smaller chunks, but this approach will increase the number of `GET` calls to the network.  
+As a workaround the `update raw data` API can be invoked in smaller chunks, but this approach will increase the number of `GET` calls to the network.
+
+# Alternatives
+To overcome the drawback, we can probably derive motivation from the REST based cloud APIs such as AWS.
+The way they handle large file uploads is bit different.
+
+1. API is invoked to indicate a beginning of an upload along with the actual content size that would be uploaded. This will return an temporary token or a request id.
+2. Then the actual upload of the content starts in smaller chunks, this request will refer to the temporary token obtained from step1.
+3. The upload is continued till the file is completely uploaded.
+
+A similar approach can be adopted as an enhancement later, without disturbing the existing API.   
+
+# Unresolved questions
+
+None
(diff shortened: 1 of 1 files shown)


proposed/0016-launcher-data_types-api/0016-launcher-data-types-api.md
```diff
@@ -0,0 +1,282 @@
...
+# Detailed design
+
+The APIs that would facilitate operations on the SafeNetwork Data types can be grouped
+under the end point `safe-api/v1.0/data_types/`.
+
+## Structured data API
+
+#### Create Structured data
+Structure Data can be created only within the permitted range as specified in the [reserved-names RFC](https://github.com/maidsafe/rfcs/blob/master/implemented/0003-reserved_names/0003-reserved_names.md#detailed-design),
+i.e within the permissible range 10,001 to 2^64
```
theWebalyst: This presumably refers to the tag_type, but it is confusing as the text refers to both "names" (ie text) and "range 10,001 to 2^64" (numeric) so it would help to be explicit.

proposed/0016-launcher-data_types-api/0016-launcher-data-types-api.md
```diff
@@ -0,0 +1,282 @@
...
+
+##### Request
+```javascript
+{
+  endpoint: 'safe-api/v1.0/data_types/create_structured_data',
+  data: {
+    id: String, // Id of the structured Data
+    tag_type: Integer,
```
theWebalyst: The "reserved-names RFC" refers to type_tag whereas below we have tag_type so one needs correcting.

proposed/0016-launcher-data_types-api/0016-launcher-data-types-api.md
```diff
@@ -0,0 +1,282 @@
...
+
+#### Request
+```javascript
+{
+  endpoint: 'safe-api/v1.0/data_types/save_raw_data',
+  data: {
+    content: String // base64 String
+  }
+}
+```
```
theWebalyst: One of the features discussed to make the REST API more useful was GET/PUT/POST of particular versions - these do nothing, but return an error if the version found is not the one referred to. 

This is a fairly common standard for REST/CRUD interfaces, because it makes collaborative working possible with just the REST API (I think - going from memory here).

I mention it (possibly not in the best place, sorry) so it doesn't get forgotten. I understand that there may be other planned ways to support collaborative use of storage, but this is worth considering if feasible (and it would seem to be straightforward?) because many existing applications would sit easily on this.

proposed/0016-launcher-data_types-api/0016-launcher-data-types-api.md
krishnaIndia: Thanks for your input @theWebalyst. You are right, it should be tag_type. I will update the RFC.

PR #82, opened 2016-01-20
Mpid messaging amendments

brian-js:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/82)

### DIFF
--- a/agreed/0009-MPID-Messaging/0009-MPID-Messaging.md
+++ b/agreed/0009-MPID-Messaging/0009-MPID-Messaging.md
@@ -56,7 +56,7 @@ pub const MAX_OUTBOX_SIZE: usize = 1 << 27;  // bytes, i.e. 128 MiB
 
 ```rust
 pub struct MpidHeader {
-    sender_name: ::NameType,
+    sender_name: XorName,
     guid: [u8; 16],
     metadata: Vec<u8>,
     signature: ::sodiumoxide::crypto::sign::Signature,
@@ -71,8 +71,8 @@ The `metadata` field allows passing arbitrary user/app data.  It must not exceed
 
 ```rust
 pub struct MpidMessage {
-    header: ::MpidHeader,
-    recipient: ::NameType,
+    header: MpidHeader,
+    recipient: XorName,
     body: Vec<u8>,
     recipient_and_body_signature: ::sodiumoxide::crypto::sign::Signature,
 }
@@ -89,7 +89,7 @@ This can be implemented as a `Vec<MpidMessage>`.
 
 Again this will be one per MPID (owner), held on the MpidManagers, and synchronised by them at churn events.
 
-This can be implemented as a `Vec<(sender_name: ::routing::NameType, sender_public_key: ::sodiumoxide::crypto::sign::PublicKey, mpid_header: MpidHeader)>` or having the headers from the same sender grouped: `Vec<(sender_name: ::routing::NameType, sender_public_key: ::sodiumoxide::crypto::sign::PublicKey, headers: Vec<mpid_header: MpidHeader>)>` (however this may incur a performance slow down when looking up for a particular mpid_header).
+This can be implemented as a `Vec<(sender_name: XorName, sender_public_key: ::sodiumoxide::crypto::sign::PublicKey, mpid_header: MpidHeader)>` or having the headers from the same sender grouped: `Vec<(sender_name: XorName, sender_public_key: ::sodiumoxide::crypto::sign::PublicKey, headers: Vec<mpid_header: MpidHeader>)>` (however this may incur a performance slow down when looking up for a particular mpid_header).
 
 ## Messaging Format Among Nodes
 
@@ -234,8 +234,8 @@ Requests composed by Client:
 | Put     | Client(A) creating a new message                                    | `Wrapper::MpidMessage`                 | Managers(A)                     |
 | Post    | Client(A) checking existence of list of sent messages in own outbox | `Wrapper::OutboxHas(Vec<Header.name>)` | Managers(A)                     |
 | Post    | Client(A) getting list of all messages still in own outbox          | `Wrapper::GetOutboxHeaders`            | Managers(A)                     |
-| Delete  | Client(A) or (B) deleting from own outbox or inbox respectively     | `NameType`                             | Managers(A) or (B) respectively |
-| Delete  | Client(B) deleting a "read" message from sender's outbox            | `NameType`                             | Managers(A)                     |
+| Delete  | Client(A) or (B) deleting from own outbox or inbox respectively     | `XorName`                              | Managers(A) or (B) respectively |
+| Delete  | Client(B) deleting a "read" message from sender's outbox            | `XorName`                              | Managers(A)                     |
 | Post    | Client announcing to Managers it's connected to network             | `Wrapper::Online`                      | Managers                        |
 
 Requests composed by Client:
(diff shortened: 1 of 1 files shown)


agreed/0009-MPID-Messaging/0009-MPID-Messaging.md
```diff
@@ -1,6 +1,6 @@
 - Feature Name: MPID Messaging System
 - Type: New Feature
-- Related Components: [safe_vault](https://github.com/maidsafe/safe_vault), [safe_client](https://github.com/maidsafe/safe_client), [routing](https://github.com/maidsafe/routing)
+- Related Components: [safe_vault](https://github.com/maidsafe/safe_vault), [safe_client](https://github.com/maidsafe/safe_client), [mpid_messaging](https://github.com/maidsafe/mpid_messaging)
 - Start Date: 22-09-2015
```
maqi: this link https://github.com/maidsafe/mpid_messaging opens nothing.
and, are we going to have a separate crate for messaging? instead of having a persona mpid_manager inside vault?

agreed/0009-MPID-Messaging/0009-MPID-Messaging.md
```diff
@@ -93,16 +93,11 @@ This can be implemented as a `Vec<(sender_name: ::routing::NameType, sender_publ
...
-    type_tag: MPID_MESSAGE,
-    identifier: mpid_message_name(mpid_message),
-    data: ::utils::encode(mpid_message),
-    previous_owner_keys: vec![],
-    version: 0,
-    current_owner_keys: vec![sender_public_key],
-    previous_owner_signatures: vec![]
+let pd = PlainData {
+    name: mpid_message_name(mpid_message),
+    data: serialise(MpidMessageWrapper::PutMessage(mpid_message)),
```
maqi: any detailed description of this MpidMessageWrapper?

agreed/0009-MPID-Messaging/0009-MPID-Messaging.md
brian-js: The repository has not been created yet, the link should work when that's done. Yes, a separate crate for messaging.

agreed/0009-MPID-Messaging/0009-MPID-Messaging.md
brian-js: Probably best to follow https://maidsafe.atlassian.net/browse/MAID-1550 and update it with whatever we need.

Fraser999: That link doesn't show `MpidMessageWrapper`.  Also, I think the detailed info should be in the RFC for completeness.

agreed/0009-MPID-Messaging/0009-MPID-Messaging.md
Fraser999: Just to clarify - that repo is really just intended to provide the message types required by the Client and Vault libs in order to avoid creating a dependency between these two libs.  The persona will still be implemented in Vaults.

agreed/0009-MPID-Messaging/0009-MPID-Messaging.md
Fraser999: @brian-js Could this also be updated to replace all instances of `NameType` with `XorName` please?

brian-js: Sure, to above.

The intention was to replace MpidMessageWrapper with those enums but looks like Qi is happy with MpidMessageWrapper so I'll update the task.

maqi: The MpidMessageWrapper I used in the implementation is quite different to the one defined in the rfc.
The rfc one is for the use of SD, which needs a typetag to differentiate the operations.
The one used in the implementation is a wrapper of data (MpidHeader or MpidMessage)

dirvine: I feel then the RFC should have been debated/updated prior to this. We cannot have RFC's going out of line with production implementations or it will wreck the process. Please make sure we implement the RFC or change the RFC

agreed/0009-MPID-Messaging/0009-MPID-Messaging.md
```diff
@@ -381,7 +371,7 @@ impl MpidMessage {
                                                      public_key) && self.header.verify(public_key)
     }
 
-    fn encode(recipient: &::NameType, body: &Vec<u8>) -> Vec<u8> {
+    fn encode(recipient: &XorName, body: &Vec<u8>) -> Vec<u8> {
         ::utils::encode(&(recipient, body)).unwrap_or(vec![])
```
maqi: Is this encode/decode util stuff still available, or it gives way to the serialise/deserialise functions?
And, shall we prefer a wrapper class to holds the fields (recipient, body), so serialise/deserialise can be used smoothly.

agreed/0009-MPID-Messaging/0009-MPID-Messaging.md
```diff
@@ -533,17 +520,12 @@ impl MpidManager {
...
-            type_tag: MPID_MESSAGE,
-            identifier: header.sender_name,
-            data: ::utils::encode(MpidMessgeWrapper::GetMessage(mpid_header_name(mpid_header))),
-            previous_owner_keys: vec![],
-            version: 0,
-            current_owner_keys: vec![header.sender_public_key],
-            previous_owner_signatures: vec![]
+        let request_sd = PlainData {
+            name: mpid_header.sender_name,
+            value: seialise(MpidMessgeWrapper::GetMessage(mpid_header_name(mpid_header))),
```
maqi: referring to the implementation PR https://github.com/maidsafe/safe_vault/pull/317
it turns out the MpidMessageWrapper only needs to be serving as content wrapper, holding MpidHeader or MpidMessage .
I remember the previous enum is to serving as a operation tag in SD.

agreed/0009-MPID-Messaging/0009-MPID-Messaging.md
```diff
@@ -337,15 +327,15 @@ pub const MAX_BODY_SIZE: usize = ::structured_data::MAX_STRUCTURED_DATA_SIZE_IN_
...
 }
 
 impl MpidMessage {
-    pub fn new(header: ::MpidHeader,
-               recipient: ::NameType,
+    pub fn new(header: MpidHeader,
+               recipient: XorName,
                body: Vec<u8>,
                secret_key: &::sodiumoxide::crypto::sign::SecretKey)
                -> Result<MpidMessage, ::error::RoutingError> {
```
maqi: so, RoutingError to be used, ? or shall define own Error?

dirvine: I think there was the notion that the shared crate for vaults and clients would define these things like data error etc. So perhaps the errors would be best done there.

agreed/0009-MPID-Messaging/0009-MPID-Messaging.md
dirvine: serialise/deserialise are in maidsafe_utils and this will work here (traits should all be satisfied).

Fraser999: I think `serialise`/`deserialise` is the preferred way now, but I don't think we need the wrapper.  The `encode` here is just to allow us to create serialised data in order to sign it - it's probably OK as shown here (just create a tuple on the fly).

agreed/0009-MPID-Messaging/0009-MPID-Messaging.md
Fraser999: +1

PR #87, opened 2016-01-28
Some more changes to Node Insertion.

afck:
The close group of a bucket address can actually only respond with
public IDs, not with endpoints, since these will in the future contain
the contact info that has to be specifically prepared for each upcoming
connection.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/87)


dirvine:
Looks good to me and as this is in active discussion I would go for merge here.

PR #99, opened 2016-02-18
Add 'crust mock' RFC

madadam:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/99)

### DIFF
--- a/proposed/0023-crust-mock/0023-crust-mock.md
+++ b/proposed/0023-crust-mock/0023-crust-mock.md
@@ -27,13 +27,13 @@ A new trait `ConnectionService` MUST be defined. It MUST have all the relevant m
 
 ```rust
 pub trait ConnectionService {
-  fn start_listening_tcp();
-  fn start_listening_utp();
-  fn start_service_discovery();
-  fn prepare_connection_info();
-  fn connect();
-  fn disconnect();
-  fn send();
+  fn start_listening_tcp(&mut self) -> io::Result<()>;
+  fn start_listening_utp(&mut self) -> io::Result<()>;
+  fn start_service_discovery(&mut self);
+  fn prepare_connection_info(&mut self, result_token: u32);
+  fn connect(&self, our_info: OurConnectionInfo, their_info: TheirConnectionInfo);
+  fn disconnect(&self, id: &PeerId) -> bool;
+  fn send(&self, id: &PeerId, data: Vec<u8>) -> io::Result<()>;
 }
 ```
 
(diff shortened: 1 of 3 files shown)


proposed/0023-crust-mock/0023-crust-mock.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Changes to crust
+
+A new trait `ConnectionService` MUST be defined. It MUST have all the relevant methods of `crust::Service`:
+
+```rust
+pub trait ConnectionService {
+  fn start_listening_tcp();
+  fn start_listening_utp();
+  fn start_service_discovery();
```
dirvine: should take a port here I think. A well known port for your service

proposed/0023-crust-mock/0023-crust-mock.md
```diff
@@ -0,0 +1,106 @@
...
+  fn prepare_connection_info();
+  fn connect();
+  fn disconnect();
+  fn send();
```
dirvine: Ah we should probably fill in all the attributes of these methods. :+1:

proposed/0023-crust-mock/0023-crust-mock.md
```diff
@@ -0,0 +1,106 @@
...
+routing::Core::new(..., |crust_event_sender| {
+  Box::new(crust::mock::Service::new(crust_event_sender))
+})
+```
+
+## Features of `crust::mock::Service`
+
+- Do not hit real network (not even LAN or localhost), do not use TCP/UDP sockets.
+- Simulate all the relevant config files crust uses (crust config, bootstrap cache, etc...). Possibly by accepting structs, representing those configs, as
+parameters to the constructor.
```
dirvine: If all three are `<Option>` Then it can satisfy this and make testing much simpler. Also allows users to not use the config_file_handler if they do not wish. For our apps it's easier to have a common paradigm like config_file_handler etc. so best of both worlds perhaps?

proposed/0023-crust-mock/0023-crust-mock.md
madadam: OK. I'll update the signatures of these functions to be the same as in `crust::Service`.

proposed/0023-crust-mock/0023-crust-mock.md
madadam: I'd love to be able to pass a `Config` object to (the real) `crust::Service`, it would make testing much easier. But that is slightly out of scope of this RFC. For the mock crust, we definitely NEED a way to pass configs to it.

dirvine:
@rossmuir  This should be moved to active as it's in place now.

rossmuir:
@dirvine > https://github.com/maidsafe/rfcs/issues/110

PR #112, opened 2016-04-05
Appended By Anyone Structured Data

krishnaIndia:
RFC details an approach through which dynamic data can be handled in the network.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/112)


proposed/0026-dynamic-data-support/0026-dynamic-data-support.md
```diff
@@ -0,0 +1,153 @@
+- Feature Name: Ability to Support Dynamic Data Handling
```
dirvine: Perhaps a better title is "append by all" structured data type as there will be several types such as whitelisted append SD etc. as well as private append only etc. the network will handle.  A small bit at a time is good though.

Issue #120, opened 2016-04-27
Sub-Directories make it hard to link to RFCs: how about using a status-field instead?


ligthyear:
I understand and totally agree that the way rust handles their RFCs is pretty unclear regrading the state of each RFC and see the reason behind creating the folders. However, this makes it hard to link to any RFC as their link throughout the development cycle changes significantly, one reason why RFCs can't properly link to each other directly within the document.

Coming from the [Python PEP Process](https://www.python.org/dev/peps/) (which the rust RFC process is largely inspired by), how about adding a top-level field `status` instead like they do (and maybe one that mentions the release version once it was published)? This way, each RFC could stay in one place and be linked to.

In regards to have a good overview, I'd suggest to do it similarly as Rust does it and just link the RFCs from the Readme.md, but as sections depending their status (ordered by number, optionally descending). Changing the status then needs to be PR which also moves them from one section to another.

dirvine:
Nice idea. @rossmuir can certainly help here. He is currently pushing the RFC process and making it more friendly. I like the status field idea as it will keep links clear and valid. @rossmuir can you please jump in and help out here. Some really good ideas and several fixes. This will help a lot with the current push for clarity and more community involvement.

rossmuir:
Sure @dirvine sounds good - I will check out the Python PEP Process - thanks @ligthyear

ligthyear:
I'm happy to make a PR with the changes (moving all RFCs into one sub-dir and adding a `status`-flag in the top, and optionally link all RFCS to one another and their appropriate github tickets), if that'd be helpful and accepted.

rossmuir:
@ligthyear that would be very helpful thanks :+1:  - sorry I have not got to this. This will improve linking and simplify the process - thank you.

rossmuir:
Thank you @ligthyear - This really does simplify and improve the process.

PR #121, opened 2016-05-02
Restructure File to use status field rather than directory structure

ligthyear:
Following up on #120 this PR adds a new "status" field on every RFCs to indicate the state ("proposed", "agreed", "active", "implemented", "rejected") of the RFC within the process. This allows us to remove the directory structure and instead put all RFCs at one place, which never changes, allowing us and others to link to them forever. This PR contains a commit per movement of set of RFCS of the same status, followed by one commit adding the status field for those just moved.

To make it easier to keep the overview, I also added a small script which looks up the RFCs by status and generates the also newly added "RFCs-by-status" Markdown file.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/121)


ligthyear:
Didn't quite realise that, but it appears my editor also automatically did some white-space fixes (and I did one typo fix) as part of this. I hope this is no problem.

rossmuir:
Thank you @ligthyear, I will review review this one.

rossmuir:
Thank you for all the effort you have put in here @ligthyear  - I shall communicate the updated process to the community via the Dev Update tonight - we owe you :beers:

PR #128, opened 2016-05-10
Added RFC 0033 - Fake Account Packet.

Fraser999:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/128)

### DIFF
--- a/text/0033-fake-account-packet/0033-fake-account-packet.md
+++ b/text/0033-fake-account-packet/0033-fake-account-packet.md
@@ -1,0 +1,206 @@
+- Feature Name: Fake Account Packet
+- Status: proposed
+- Type: enhancement
+- Related components: SAFE Vault and Routing
+- Start Date: 10-05-2016
+- RFC PR:
+- Issue number:
+
+# Summary
+
+This proposes an enhancement whereby the network will respond to a `Get` request for a non-existent
+account packet (`StructuredData` with `type_tag` 0) with a fake account packet.
+
+# Motivation
+
+The objective is to deter an attack where random account packets are requested by a malicious Client
+in the hopes of finding one in order to attempt decryption of it.
+
+The defence is to have the network respond to requests for non-existent account packets with fakes
+which will appear valid to the recipient.  This will make it expensive for the attacker, since it
+won't know whether the decrypted packet will yield useful data or not.
+
+# Detailed design
+
+## Issues to Overcome
+
+The main issue is to be able to have a group of Vaults agree a fake packet which will be accumulated
+by the attacker - i.e. the group all need to generate an identical fake.  Essentially we just need a
+single value agreed across the group which can then be used to seed a pseudorandom number generator.
+
+The Vaults could all exchange a message in order to agree a seed, however this is undesirable since
+it could easily be exploited as an amplification attack.
+
+This leaves little scope for agreeing a single seed; the only piece of information which all the
+Vaults know is the list of Vault names comprising the close group.  This is useless as a seed since
+the attacker could simply monitor the individual Routing responses arriving and hence accrue this
+list of Vault names.  It would then be easy to deduce that the responses contain a fake.
+
+A more trivial issue is that while account packets are currently the highest-value targets for an
+attacker, it should be possible to extend the defence to other `StructuredData` types easily.
+
+## Proposed Solution
+
+There is an example implementation at https://gitlab.com/Fraser999/Fake-Account-Packet.
+
+The implementation here requires Vaults to hold a new piece of data; a [`SessionId`]
+(https://gitlab.com/Fraser999/Fake-Account-Packet/blob/master/src/main.rs#L21) defined as
+
+```rust
+struct SessionId(pub [u8; SEEDBYTES]);
+```
+
+where `SEEDBYTES` is 32; the size of the sodiumoxide `Seed`.
+
+This `SessionId` is created randomly by the Vault on startup and is exchanged with any peers with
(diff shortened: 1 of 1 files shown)


text/0033-fake-account-packet/0033-fake-account-packet.md
```diff
@@ -0,0 +1,206 @@
...
+where `SEEDBYTES` is 32; the size of the sodiumoxide `Seed`.
+
+This `SessionId` is created randomly by the Vault on startup and is exchanged with any peers with
+which it establishes a connection.  Critically, this `SessionId` is never sent to any node (Vault or
+Client) until the peer is being added to the Routing Table.  This should reduce the ability for an
+attacker to harvest `SessionId`s.
+
+This new piece of information per Vault (unknown by Clients) can now be used to generate a seed for
+any given close group.  To avoid generating the same seed for all addressable elements within a
+narrow address space it would be best to also mix in the requested data name.
```
dirvine: I think it may be necessary to hash the collected seeds with request `name` or similar. As a one way function then it becomes very difficult (or maybe impossible) to know the attacker is getting fake packets.

text/0033-fake-account-packet/0033-fake-account-packet.md
```diff
@@ -0,0 +1,206 @@
...
+                        Some(&private_key))
+        .expect("Can't fail.")
+}
+```
+
+# Drawbacks
+
+This presents a very basic implementation.  It requires the Vaults to generate throwaway
+cryptographic keys in order to create a plausible fake - a relatively expensive operation.  For an
+attacker just looking to do a DOS attack, this might provide a vector.  Profiling may show that this
```
dirvine: IF there is an agreed list of seeds then the hash mentioned above would perhaps alleviate this attack.

text/0033-fake-account-packet/0033-fake-account-packet.md
Fraser999: I'm not sure I follow.  The seeds aren't transmitted to the client - there's no feasible way for the attacker to know what seeds were used as far as I can see.

I don't think we benefit here using hashing - but maybe I'm missing something?

text/0033-fake-account-packet/0033-fake-account-packet.md
Fraser999: Again, I don't think we're on the same page.  Even if we were to hash data names and/or session IDs, we'd still need the peers to create throwaway crypto keys I think?

text/0033-fake-account-packet/0033-fake-account-packet.md
dirvine: What I mean is hash the seed with the hash of the original message (or similar)  of the `Get` and everyone does the same. so we have one way function to provide a fake reply.

afck:
I haven't fully thought that through, but an alternative might be to use the actual data chunk (immutable or structured - it doesn't matter) that's closest to the requested (nonexistent) one as a seed. That information - which one is the closest existing chunk to that address - should also be known by only the close group of that data, and as soon as there are much more data chunks than nodes, the close group of a nonexistent data name will almost always be identical with the close group of a close existing chunk, so their responses will accumulate.

text/0033-fake-account-packet/0033-fake-account-packet.md
```diff
@@ -0,0 +1,206 @@
...
+timeout).
+
+2. Respond with `GetFailure` for non-existent data, but monitor the rate at which such requests are
+passing by.  This would need to be handled by the Vault through which the Client is connected.  Once
+an unacceptably high rate of `GetFailure`s is reached, the Client is dropped.  This seems fairly
+simple to implement, but an attacker would be able to work around this by using many Clients.
+
+# Unresolved questions
+
+1. Should the seed be generated using all members' names from the close group, or just a subset of
```
dirvine: routing should give the correct close group, unless churn is happening. This churn can affect even a smaller set, but may be alleviated again.

Fraser999:
@afck I think as well as the initial possibility for discrepancies when the network is sparsely populated with chunks, there's also maybe scope for synchronisation issues to cause problems - where not all members of the group agree on the closest chunk since it's part way through being stored or deleted.

Having said that - for all the times that'll happen, all we lose is that the attacker gets to know that that particular response is a fake.  As long as it happens very infrequently, there shouldn't be a problem.

I can add this as the favoured approach (since it should be the easiest to implement) if no-one else objects.

text/0033-fake-account-packet/0033-fake-account-packet.md
Fraser999: Yup - agreed.  This would only reduce the chance of churn affecting the seed, but not eliminate it.  And in theory we could as equally well choose the furthest members of the close group and achieve the same level of resilience to churn.

I'm happy to leave it as the whole group being used though, if there are no objections to that.

text/0033-fake-account-packet/0033-fake-account-packet.md
Fraser999: Not sure.

So what I'm proposing is that we create a seed which is derived from the 8 session IDs and the data name.  The session IDs shouldn't be known to the attacker unless he's connected to the entire group.  If he *is* connected to the group, he already knows of any valid account packets - they'll have been refreshed to him.

If he has somehow managed to find the session IDs another way then hashing won't help since the sender (attacker) knows what the hash value is, since he sent the message.

Mixing in the data name is purely to generate different seeds for each different address managed by a single group.

Issue #124, opened 2016-05-03
RFCs-by-status.md is messed up around the Launcher stuff


theWebalyst:
The following two links have the same link text "#0018 Launcher as a local server"

https://github.com/maidsafe/rfcs/blob/master/text/0018-launcher-as-rest-server/0018-launcher-as-local-rest-server.md
https://github.com/maidsafe/rfcs/blob/master/text/0018-launcher-as-rest-server/authorisation_api.md

And the links to the dns_api.md and nfs_api.md are missing.

ligthyear:
Yeah, I saw that initially, too. But I didn't want to decide on how to handle this. I could have made an exception in the script, which generates it, but it feels like a more structural decision that needs to be taken, which is: how should multi-document-RFCs be done and handled.

In the original RFCs procedures of Rust and Python, every RFC is exactly one document (as in "one file"/resource). We already kinda broke this by having additional graphs and explanatory images, hence the folders for each RFC. But additionally some folders contain multiple documents (most often divided by the components they touch), some with an Index-Like-Document of the same name as the RFC itself, some not. Some with meta-headers for each file (which is what you are seeing for `#0018`), some without any. I'd say, we need to define and decide upon one structure that and fix all others to comply to that so we can solve this issue once and for all.

I'd personally argue for an entry-point file of the exact same name, which has the metadata and additional files in that folder without that metadata (but maybe a short line in the top, which RFC they belong to). But in that case, I'd also replace the feature-header with a `#-marked headline in the first line.

theWebalyst:
@ligthyear:
> I'd personally argue for an entry-point file of the exact same name, which has the metadata and additional files in that folder without that metadata (but maybe a short line in the top, which RFC they belong to). But in that case, I'd also replace the feature-header with a `#-marked headline in the first line.

This seems sensible and straightforward so +1, and I'd like to add my appreciation, thank you for improving the accessibility of these documents. Super job! :-)

PR #138, opened 2016-05-19
Unify RFC Naming and Structure

ligthyear:
In reference to #124 I restructured the RFCs to have unified naming scheme and title cases for each file. It also updates the Readme and template to explain
that schema.

Further more this PR contains an updated version of the generator script, which thanks to the new scheme will print each RFC at most once, as well as the updated result of that script.

However, this does not yet totally fix issue #124 as RFC #0024 still doesn't show up. That would require a rewrite of that RFC (and potentially merge it with #20?), for which I am missing the background information on how that should be done.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/138)


rossmuir:
Thanks very much for the PR - it helps a lot.

PR #151, opened 2016-07-03
Headings

iancoleman:
The readme in the root of the rfc repo says

The first line "should be the only first-degree headline in the entire RFC!"

There were a lot of other first-degree headlines so I changed the depth in by one for all headings that were not on the first line.

This was not possible for rfcs which were already utilizing the maximum headline depth, namely 0016 0018 0028 and 0036.

I know this pull request seems like 'busy work' but correct headings is

* the correct procedure
* better for SEO.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/151)

### DIFF
--- a/0000-template.md
+++ b/0000-template.md
@@ -8,28 +8,28 @@
 - Supersedes: (fill me in with a link to RFC this supersedes - if applicable)
 - Superseded by: (fill me in with a link to RFC this is superseded by - if applicable)
 
-# Summary
+## Summary
 
 One paragraph explanation of the feature.
 
-# Motivation
+## Motivation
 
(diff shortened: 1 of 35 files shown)


dirvine:
Thank you very much, this is not busy work, but necessary work. Very much appreciated. 

---

Reviewed 3 of 3 files at r1, 32 of 32 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/rfcs/151#-:-KLlGC7s3GyyLjw4xN2e:31762854)*

PR #153, opened 2016-07-05
Vault RPC

iancoleman:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/153)

### DIFF
--- a/text/0000-vault-rpc/0000-vault-rpc.md
+++ b/text/0000-vault-rpc/0000-vault-rpc.md
@@ -1,0 +1,210 @@
+# Vault RPC
+
+- Status: proposed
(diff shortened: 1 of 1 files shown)


dirvine:
Nice, for sure should be considered. (of course issue for tracking will be created) 

I wonder if 

1.  The security of data could be via signed messages by the private key of the wallet address.
2.  A subset of non security threatening stats could be via non encrypted/signed calls to allow nodes to subscribe to a centralised dashboard (opt in for trace nodes on the network).

iancoleman:
Seems like this feature will happen inevitably so I'll close this PR in anticipation of a maidsafe implementation in the future.

PR #159, opened 2016-07-20
doc/rfc29 Data chains RFC

dirvine:
r? @Viv-Rajkumar

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/159)


vinipsmaker:
I didn't see a discussion link, then I'm going to post here.

> A [Block] is made up of a [BlockIdentifier] and a vector of `PublicKey` and `Signature`.

So both are published and shared among the network nodes.

I've read some discussions from the bitcoin community when it relates to "quantum computing vulnerability". [The discussions usually state something in these lines](http://bitcoin.stackexchange.com/questions/6062/what-effects-would-a-scalable-quantum-computer-have-on-bitcoin):

> With a quantum computer, you could easily deduce the private key corresponding to a public key

I haven't read enough to understand how that works or even to check if all public key encryptions are affected. However I thought I should bring this up to be considered (if it hasn't been already).

andrebeat:
> An interesting aspect though is the ability to "validate in history". 

This is really interesting!

I'm not sure how deletion is handled though, since I can remove a block from my chain but other nodes can republish it and it would still be considered valid.

The fact that `StructuredData` is versioned isn't really mentioned on the RFC (but exists in the implementation). I think it should be mentioned since that's what makes updates work. I also think that it should indicate that there are no ordering guarantees (within links, or maybe even across links?).

andrebeat:
@vinipsmaker There's an algorithm for a theoretical quantum computer which can do integer factorization in polynomial time (https://en.wikipedia.org/wiki/Shor%27s_algorithm). This would break public-key encryption since it relies on factorization being a *really hard* problem. I'm not sure this is something that should be considered for now though.

dirvine:
@andrebeat  Yes that is very true. Ledger types are not mentioned here, nor is the fact for now a node can maliciously remove blocks (but not links). My thoughts here atm are 

1. We need another RFC (rather simple) to handle ledger based SD. In that case every update (ledger type) will require to be Put (as it's adding to network). Post will be restricted to updating non ledger based SD. The cost to Put respects the network is doing more work (adding something and maintaining it) 

2. Ordering (loose) and block removal are probably both solved by updates to this design. We can identify the optimum space for each data block (by relocating them on validate == true) and when ordered like that on next churn event, Nodes can sign the ordered block. There are a few ways to manage this (special block, sign each block as successor etc.). This also allow chains to be eventually ordered (similar to eventual consistency) as that is likely a valid approach in decentralised systems. 

We also consider that a node removing blocks causes itself harm as when it tries to reconnect, it's chin is shorter than other chain holders. So therefore loses its ability to restart in old position. That means starting over collecting a lot of data again. The other nodes accumulate and will catch a node in the act if on line, so on restart is when blocks can be removed. During session it should be detected and killed off. 

Good to know you are diving in, the more eyes here the better. Not looking for perfection straight away (we will not ignore security tho), but perhaps we could add a todo list to the README.md to not miss these poitns. Maybe github issues would work as well mind you. 

Thanks again

andrebeat:
1. Yes, I noticed that the ledger type is already implemented.
2. That seems like a good approach, the ordering would be enforced on each churn event / link creation.

I didn't think of the deterrent network effect of "forging" a chain, great 👍. 
As long as there are solutions in sight for these problems (which is the case) I agree that these can be solved later on. Now that I think of it, both these issues are, in a way, already present in the RFC under *Unresolved Questions*. But I agree that they could be made more explicit (maybe create github issues).

PR #165, opened 2016-08-04
Deletable Immutable-Data

ustulation:
This RFC provides motivation and approach for deletable ImmutableData.

[Rendered](https://github.com/ustulation/rfcs/blob/deletable-immutable-data/text/0000-deletable-immutable-data/0000-deletable-immutable-data.md)

---

This change is [Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/165)


text/0000-deletable-immutable-data/0000-deletable-immutable-data.md
```diff
@@ -0,0 +1,68 @@
...
...the `name` of `ImmutableData`) is derived from the hash of its content (also referred to as the `body`). Any change in the content will consequently produce a new name and the network will store it in a different place. Also `ImmutableData` is de-duplicated, i.e. if several users upload the same `ImmutableData`, they will all be charged but the network itself will have only one primary copy of it. However there are several cases where de-duplication is very unlikely, for instance if the content was encrypted or if the content was a representation of versioned directory listing of a particular user etc. In these cases when the user updates the content, they might no longer have any need for the previous content. There are also cases where one would like to delete a previously created `ImmutableData`. Currently there is neither a mechanism nor an incentive to delete unneeded `ImmutableData`. In practice we find such accumulation of never required data highly wasteful. Each churn event must account for exchange of all such data too as the network has no knowledge of what is required and what is not. Being able to delete `ImmutableData` would be nice to have and should be incentivised.
```
dirvine: perhaps 
de-duplication is very unlikely, for instance if the content was encrypted
to
de-duplication is very unlikely, for instance if the content was privately encrypted

ustulation: Wouldn't _encrypted_ signify privacy ?
`privately encrypted` seems to suggest that it could be `publicly encrypted` too which would be a vague statement.

dirvine: Yes mostly, but convergent encryption or what we do with self encryption is not private from others with the same data, it's deterministic, albeit based on having the exact same input data.

ustulation: Ah right.
Also one more (though remote) possibility for fun: if `A`, `B`, `C` etc created exactly the same data and wanted only `Z` to read they all could have used `Z's` public key for asymmetric encryption. This is also a case where an  encrypted data would be de-duplicated. :)

Updated.

text/0000-deletable-immutable-data/0000-deletable-immutable-data.md
```diff
@@ -0,0 +1,68 @@
...
+
+There is however a downside - for popular data which might be deduplicated a large number of times, `Owner::Users` will grow enormously in size. A solution to this is that deletion of `ImmutableData` must be entail a _best-effort_ guarantee. If the deduplication rises above a limit (hence highly popular), we will consider that data to be entirely network owned and can no longer be deleted by any of its creators. At this point the storage will flip `owner` to `Owner::Network` and beyond this, `ImmutableData` can no longer be deleted and has no associated owners. The policy of reimbursement will not be applicable anymore.
+
+This upper limit shall be called `DEDUPLICATION_LIMIT` and shall be defined as follows: _The number of unique keys in `Owner::Users` beyond which this `ImmutableData` is entirely owned by the network_.
+Its value is chosen thus:
+```rust
+const DEDUPLICATION_LIMIT: usize = 20;
```
dirvine: Another nice feature of this is that a potential vandal trying to fill the network with rubbish that does not delete would need to pay for it many time over (20 here). 

Of course it does not answer the other question about a vandal uploading stuff that does not get deleted (i.e. they never call delete anyway, since they are attempting vandalism)

text/0000-deletable-immutable-data/0000-deletable-immutable-data.md
```diff
@@ -0,0 +1,68 @@
...
+```
+Once number of unique keys in `Owner::Users` is greater than `DEDUPLICATION_LIMIT`, an attempt to delete such an `ImmutableData` shall result in the following (existing) error:
+```rust
+// Defined in safe_network_common at the time of this RFC
+MutationError::InvalidOperation
+```
+
+## Drawbacks
+- A slight increase in complexity at vault level for handling of `ImmutableData`.
+- The deletion (and hence the incentive of reimbursement) is not a full guarantee but a best-effort one.
```
dirvine: I think another drawback may be churn. these lists will require to merge in cases where users are being added/removed from the list. During churn this will need to accumulate all the not yet locked lists.

ustulation: updated

text/0000-deletable-immutable-data/0000-deletable-immutable-data.md
ustulation: > fill the network with rubbish that does not delete would need to pay for it many time over (20 here).
Yes and that too from different accounts as the definition entails _unique_ keys.

text/0000-deletable-immutable-data/0000-deletable-immutable-data.md
```diff
@@ -0,0 +1,70 @@
...
+
+//Conceptual storage in a node:
+let storage: HashMap<XorName, ImmutableData>;
+```
+For someone to be able to delete it there has to be some notion of ownership tied to it, otherwise `B` can delete data that was created only by `A`. We need a mechanism in place where `B` can delete data created by `A` iff `B` too had created exactly the same data. After `B` has done deleting, the data should still remain as `A` hasn't deleted it yet. Incentivising would mean that whoever deletes data gets reimbursed for playing their part in freeing up space. From this point of view, we don't want `B's` subsequent delete attempts to be legit (allowed), because `B` has already been reimbursed. When `A` finally issues a delete, a reimbursement is credited and data is finally deleted by the network.
+
+Changing what is mentioned as a conceptual storage of `ImmutableData` in a node above can help achive this:
+```rust
+enum Owner {
+    Users(Vec<sign::PublicKey>),
```
Viv-Rajkumar: Why not have this `Users(HashMap<sign::PublicKey, u64>) // u64 - for ref count`. Would it help with amount of data transferred curing churn?

dirvine: Certainly easier to get the unions of agreed number of keys (accumulate) I think

ustulation: updated

Viv-Rajkumar:
Another advantage might be Delete requests not needing requestor to send the data as part of `Delete` request RPC.

- **ImmutData** Just the requestor's SignedMessage key should be enough to validate ownership and process the request.
- **StructuredData**(Future scope) Have delete only succeed when there remains a single owner at which point the same route as ImmutData can apply, thereby making ownership transfers/changes even for multi-sig potentially restricted to POST requests

Maybe some future scope should get detailed too which further RFCs should build up on.
- Provide incentives to encourage users to actually "delete" data and not just "clear their storage hierarchy"

dirvine:
Agreed, for now this uses the users key (which is not great). We have many options though, but for later RFC's.

ustulation:
There is one problem here though: If `A` has created something 5 times and deletes once, i can capture that delete and replay it multiple times until his data is completely gone from the network.

dirvine:
Not sure, you would only be able to remove that ID from the Vec/map and not others, so should still be safe.

ustulation:
But that is a potential problem.
1. What if that data was only owned by `A` who created it 5 times ?
2. Even for data owned by many, `A` is not going to be happy about this - this is a potential (and easy) attack vector.

Viv-Rajkumar:
well not related to this RFC, but for any request, replay shouldn't work considering the `MessageId` for the original request is going to make the next replay get filtered by the group via the `signed_msg_filter`, you still can wait for the timeouts to expire and then replay once and repeat same to get the original intent but thats something that in general needs to get addressed at routing/crust

ustulation:
Ok - but right now this _is_ a big problem in this rfc. Timeouts are not sufficient deterrent to replay attacks.

Viv-Rajkumar:
> I mean it's something outside the scope of this RFC as one can do the same for PUT's right now and cap an account given enough(really really long) time

Thierry61:
The possibility to delete an immutable data should be optional and reserved for the case the user wants to delete the data later, for example a temporary file. This shouldn't be the default case for 2 reasons:

- prevent accidental destruction (for example for a file stored in several directories, the user might delete the last instance without being aware of this)

- having no link between the data and the uploader is the only way to be sure of anonymity

To implement this, I propose that the PUT command is augmented by an optional public signing key. If none is provided then the immutable data is directly set in the network owner mode, without waiting for the DEDUPLICATION_LIMIT threshold.

dirvine:
I think this makes a lot of sense. Also a mechanism that ensures if a public data-map / directory is stored then perhaps there is a case that no matter what, that data is maintained forever.

ustulation:
Superseded by [unpublished-immut-data rfc](https://github.com/maidsafe/rfcs/blob/89c702efb62240f65f6990c649e2cbe531b3ddcb/text/0055-unpublished-immutable-data/0055-unpublished-immutable-data.md)

PR #166, opened 2016-08-09
doc/rfc37: add more examples

afck:
[Rendered](https://github.com/afck/rfcs/blob/master/text/0037-disjoint-groups/0037-disjoint-groups.md)

This change is [Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/166)

### DIFF
--- a/text/0037-disjoint-groups/0037-disjoint-groups.md
+++ b/text/0037-disjoint-groups/0037-disjoint-groups.md
@@ -42,29 +42,44 @@ The current group definition would considerably complicate [data chains](https:/
 
 ## Detailed design
 
-The definition of a close group is modified so that at each point in time, the name space is partitioned into disjoint groups. Each group is defined by a _name prefix_, i.e. a sequence of between 0 and `XOR_NAME_BITS` (currently 256) bits, similar to how IP subnets correspond to IP address prefixes. The group `(p)` consists of all nodes whose name begins with the prefix `p`, and is responsible for all data items whose name begins with `p`. The group thus manages the part of the name space given by the interval `[p00...00, p11...11]`. The routing table needs to keep track of not only which peers a node is connected to, but also how they are grouped.
+The definition of a close group is modified so that at each point in time, the name space is partitioned into disjoint groups. That means that every address in the network at every point in time belongs to _exactly one_ group. Each group is defined by a _name prefix_, i.e. a sequence of between 0 and `XOR_NAME_BITS` (currently 256) bits, similar to how IP subnets correspond to IP address prefixes. The group `G(p)` consists of all nodes whose name begins with the prefix `p`, and is responsible for all data items whose name begins with `p`. The group thus manages the part of the name space given by the interval `[p00...00, p11...11]`.
 
-When the network is bootstrapped, there is only one group, with the empty prefix `()`, responsible for the whole name space.
+The routing table needs to keep track of not only which peers a node is connected to, but also how they are grouped. And the network's structure is defined not by the current set of nodes alone, but in addition by the groups which currently exist. An address belongs to _exactly one group_ if and only if _exactly one_ of the address' prefixes is the prefix of a current group. So to define a partition of the name space:
 
-* Whenever `(p0)` and `(p1)` satisfy certain group requirements (see below), a group `(p)` splits into `(p0)` and `(p1)`.
-* Whenever a group `(p0)` or `(p1)` ceases to satisfy the group requirements, it merges with all its sister groups into the group `(p)` again: All groups that would be a subset of `(p)` are merged back into `(p)`.
+* No two groups must be comparable: If `G(p)` and `G(q)` are different groups, then `p` and `q` cannot be a prefix of each other - they must differ in at least one bit that is defined in both of them.
+* Every address must have a prefix that belongs to a group.
 
-These are the only two operations allowed on groups, so it is guaranteed that every address in the network belongs to exactly one group, and that group satisfies the requirements. Note that the second rule is applied as soon as _at least one_ of the groups does not satisfy the requirements anymore. E.g. if there are groups `(111)`, `(1100)` and `(1101)`, and `(111)` does not satisfy the requirements, then even if the other two do, the three groups merge into `(11)`.
+For example, these sets of groups form valid partitions of the address space:
+
+* `G(00), G(01), G(10), G(11)`
+* `G(0), G(10), G(110), G(1110), G(1111)`
+* `G()`
+
+Whereas `G(0), G(00), G(10), G(01)` would not be valid, because `0` and `00` are comparable. And `G(01), G(10), G(11)` would not be valid because an address starting with `00` belongs to none of those groups.
+
+When the network is bootstrapped, there is only one group, with the empty prefix `G()`, responsible for the whole name space.
+
+* Whenever `G(p0)` and `G(p1)` satisfy certain group requirements (see below), a group `G(p)` splits into `G(p0)` and `G(p1)`.
+* Whenever a group `G(p0)` or `G(p1)` ceases to satisfy the group requirements, it merges with all its sister groups into the group `G(p)` again: All groups that would be a subset of `G(p)` are merged back into `G(p)`.
+
+These are the only two operations allowed on groups, so it is guaranteed that every address in the network belongs to exactly one group, and that group satisfies the requirements. Note that the second rule is applied as soon as _at least one_ of the groups does not satisfy the requirements anymore. E.g. if there are groups `G(111)`, `G(1100)` and `G(1101)`, and `G(111)` does not satisfy the requirements, then even if the other two do, the three groups merge into `G(11)`.
 
 The invariant that needs to be satisfied by the routing table is modified accordingly:
 
-1. A node must have its complete group `(p)` in its routing table.
-2. It must have every member of every group `(q)` in its routing table, for which `p` and `q` differ in exactly one bit.
+1. A node must have its complete group `G(p)` in its routing table.
+2. It must have every member of every group `G(q)` in its routing table, for which `p` and `q` differ in exactly one bit.
 
-If `p` and `q` are not of equal length, that does not count as a difference: A differing bit is one that is defined in _both_ prefixes, but is 1 in one of them and 0 in the other. So e. g. `111`, `1100` and `1101` all differ in exactly one bit from each other.
+If `p` and `q` are not of equal length, that does not count as a difference: A differing bit is one that is defined in _both_ prefixes, but is `1` in one of them and `0` in the other. So e. g. `111`, `1100` and `1101` all differ in exactly one bit from each other.
 
-The groups that differ in the `i`-th bit are the "`i`-th bucket" of `(p)`.
+The groups that differ in the `i`-th bit are the "`i`-th bucket" of `G(p)`.
 
-In a balanced network, point 2 just means that the group `(q)` _is_ the `i`-th bucket. But if it is not perfectly balanced, it might be the two groups `(q0)` and `(q1)` or even more groups with longer prefixes, or some prefix of `q` itself.
+In a balanced network, point 2 just means that the group `G(q)` _is_ the `i`-th bucket. But if it is not perfectly balanced, it might be the two groups `G(q0)` and `G(q1)` or even more groups with longer prefixes, or some prefix of `q` itself.
 
 The requirement is symmetric: I need you in my routing table if and only if you need me in yours!
 
 It also implies the current Kademlia invariant: I am still connected to each of my bucket groups. (The `GROUP_SIZE` nodes closest to my `i`-th bucket address are necessarily in the same group as the bucket address itself, and therefore cannot differ in more than one bit from my own.)
+
+As an example, consider the group `G(0101)`. Its `0`th bucket consists of all groups that differ exactly in the `0`th bit from `0101`, i.e. all groups whose prefix is comparable to `1101`. That might be the group `G(11)`, the group `G(1101)` or _all three of the groups_ `G(11010)`, `G(110110)` and `G(110111)`. Similarly, its `1`st bucket consists of all groups whose prefixes are comparable with `0001` (they differ in exactly the `1`st bit from `0101`). If, for example, that happens to be exactly group `G(000)`, then the `1`st bucket must contain _all_ members of `G(000)` - those whose name starts with `0001` _and_ those whose name starts with `0000`.
 
 ### Group requirements
 
@@ -76,7 +91,7 @@ To avoid repeated splitting and merging due to small fluctuations in group size,
 
 ### Message routing
 
-To relay a message from an individual node on a given `route` in a node `n` in group `(p)` for a destination _node_ `d`:
+To relay a message from an individual node on a given `route` in a node `n` in group `G(p)` for a destination _node_ `d`:
 
 * If `d == n`, handle the message.
 * If `d` is in our routing table, relay it directly to `d`.
@@ -85,7 +100,7 @@ To relay a message from an individual node on a given `route` in a node `n` in g
 
 If the destination is a _group authority_ with address `d`:
 
-* If `p` is a prefix of `d`, handle the message and relay it to everyone else in `(p)`.
+* If `p` is a prefix of `d`, handle the message and relay it to everyone else in `G(p)`.
 * Otherwise relay the message to the `route`-th closest entry to `d` in our routing table.
 
 For any groups `A` and `B`, either everyone in `A` is closer to `d` than everyone in `B`, or vice versa. Therefore, the _group_ that we relay the message to doesn't depend on the route. Since everyone in our group knows everyone in the next hop's group, this means that in the next attempt `route + 1`, a _different_ member of `B` will receive the message. Unless there is churn in between the attempts, this guarantees that all routes are disjoint.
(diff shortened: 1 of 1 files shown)


maqi:
Review status: 0 of 1 files reviewed at latest revision, 5 unresolved discussions.

---

*[text/0037-disjoint-groups/0037-disjoint-groups.md, line 45 \[r1\]](https://reviewable.io:443/reviews/maidsafe/rfcs/166#-KOiap0qj5f5_8bi7aRk:-KOiap0rmSbVUjGb6BxT:b-ewk58w) ([raw file](https://github.com/maidsafe/rfcs/blob/9c6abfed32d68e2bee0300090bf80a66a8254f46/text/0037-disjoint-groups/0037-disjoint-groups.md#L45)):*
> ```Markdown
> ## Detailed design
> 
> The definition of a close group is modified so that at each point in time, the name space is partitioned into disjoint groups. That means that every address in the network at every point in time belongs to _exactly one_ group. Each group is defined by a _name prefix_, i.e. a sequence of between 0 and `XOR_NAME_BITS` (currently 256) bits, similar to how IP subnets correspond to IP address prefixes. The group `G(p)` consists of all nodes whose name begins with the prefix `p`, and is responsible for all data items whose name begins with `p`. The group thus manages the part of the name space given by the interval `[p00...00, p11...11]`.
> ```

we better make it clear that such `group` is from each node's perspective.
i.e. node 1 will be in node 2's G(00) but in node 3's G(01)

---

*[text/0037-disjoint-groups/0037-disjoint-groups.md, line 47 \[r1\]](https://reviewable.io:443/reviews/maidsafe/rfcs/166#-KOibADU3VVqrJ9vXzm6:-KOibADU3VVqrJ9vXzm7:bnd4jhb) ([raw file](https://github.com/maidsafe/rfcs/blob/9c6abfed32d68e2bee0300090bf80a66a8254f46/text/0037-disjoint-groups/0037-disjoint-groups.md#L47)):*
> ```Markdown
> The definition of a close group is modified so that at each point in time, the name space is partitioned into disjoint groups. That means that every address in the network at every point in time belongs to _exactly one_ group. Each group is defined by a _name prefix_, i.e. a sequence of between 0 and `XOR_NAME_BITS` (currently 256) bits, similar to how IP subnets correspond to IP address prefixes. The group `G(p)` consists of all nodes whose name begins with the prefix `p`, and is responsible for all data items whose name begins with `p`. The group thus manages the part of the name space given by the interval `[p00...00, p11...11]`.
> 
> The routing table needs to keep track of not only which peers a node is connected to, but also how they are grouped. And the network's structure is defined not by only the nodes that are currently part of it, but also by the groups which currently exist. An address belongs to _exactly one group_ if and only if _exactly one_ of the address' prefixes is the prefix of a current group. So to define a partition of the name space:
> ```

`not by only` -> `not only by` ?

---

*[text/0037-disjoint-groups/0037-disjoint-groups.md, line 82 \[r1\]](https://reviewable.io:443/reviews/maidsafe/rfcs/166#-KOicdHz-gsTIt90Z2fn:-KOicdHz-gsTIt90Z2fo:b-cw51g7) ([raw file](https://github.com/maidsafe/rfcs/blob/9c6abfed32d68e2bee0300090bf80a66a8254f46/text/0037-disjoint-groups/0037-disjoint-groups.md#L82)):*
> ```Markdown
> It also implies the current Kademlia invariant: I am still connected to each of my bucket groups. (The `GROUP_SIZE` nodes closest to my `i`-th bucket address are necessarily in the same group as the bucket address itself, and therefore cannot differ in more than one bit from my own.)
> 
> As an example, consider the group `G(0101)`. Its `0`th bucket consists of all groups that differ exactly in the `0`th bit from `0101`, i.e. all groups whose prefix is comparable to `1101`. That might be the group `G(11)`, the group `G(1101)` or _all three of the groups_ `G(11010)`, `G(110110)` and `G(110111)`. Similarly, its `1`st bucket consists of all groups whose prefixes are comparable with `0001` (they differ in exactly the `1`st bit from `0101`). If, for example, that happens to be exactly group `G(000)`, then the `1`st bucket must contain _all_ members of `G(000)` - those whose name starts with `0001` _and_ those whose name starts with `0000`.
> ```

this example is still too vague from the implementation perspective :( 
we need to explain there is `dynamic group filter (prefix)` which is decided by the max bucket len. However, I think `bucket` is a concept for implementation, and not sure whether we shall introduce it in this RFC.
but without explain it, there is a huge gap between the implementation and this rfc.

consider the group `G(0101)` -> consider node N(0101)

Its `0`th bucket consists of all groups that differ exactly in the `0`th bit from `0101`, i.e. all groups whose prefix is comparable to `1101` -> Its `0`th bucket consists of all groups that differ exactly in the `0`th bit from `0101`, i.e. all groups whose nodes' name starts with `1`  ?

---

*[text/0037-disjoint-groups/0037-disjoint-groups.md, line 135 \[r1\]](https://reviewable.io:443/reviews/maidsafe/rfcs/166#-KOieuOkLg9V0nAbvfHX:-KOieuOlGmvQ7_oXzQBj:b4edqa) ([raw file](https://github.com/maidsafe/rfcs/blob/9c6abfed32d68e2bee0300090bf80a66a8254f46/text/0037-disjoint-groups/0037-disjoint-groups.md#L135)):*
> ```Markdown
> If a group `G(p)` splits, all its nodes stay connected: The new groups are now each other's `i`-th bucket, where `i` is the number of bits in `p`.
> 
> The group sends a direct `GroupSplit(p)` message to all its contacts to inform them about the split. These need to note that the group has split, even if they stay connected to all its members. They now know that there is no group `G(p)` anymore, but there are groups `G(p0)` and `G(p1)`.
> ```

Does it mean `the node needs to send a direct `GroupSplit(p)` message to all its contacts in its routing_table` ?

---

*[text/0037-disjoint-groups/0037-disjoint-groups.md, line 157 \[r1\]](https://reviewable.io:443/reviews/maidsafe/rfcs/166#-KOif7mdHdcorIqcp46u:-KOif7meIliiIccba6Td:bgtkfzq) ([raw file](https://github.com/maidsafe/rfcs/blob/9c6abfed32d68e2bee0300090bf80a66a8254f46/text/0037-disjoint-groups/0037-disjoint-groups.md#L157)):*
> ```Markdown
> * The group initiating the merge has only one more bit than `p` and already knows all constituent groups. It sends a `GroupMerge(p, Vec<(Prefix, Vec<XorName>)>)` group message to all of them (including itself), containing the new group prefix and all its routing table entries (including itself) and _their_ group prefixes.
> * Every node accumulating a `GroupMerge` establishes connections to all entries of the list and sends a `GroupMerge(p, Vec<(Prefix, Vec<XorName>)>)` group message to all of them, in the same way.
> * Each node that has accumulated all the `GroupMerge` messages from all constituent groups of `G(p)` updates its routing table to reflect the change: It now knows that there exists a group `G(p)` and that all other groups `G(p...)` ceased to exist.
> ```

we better includes an example of split and merge, to illustrate how it works

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/rfcs/166)*

afck:
Review status: 0 of 1 files reviewed at latest revision, 5 unresolved discussions.

---

*[text/0037-disjoint-groups/0037-disjoint-groups.md, line 45 \[r1\]](https://reviewable.io:443/reviews/maidsafe/rfcs/166#-KOiap0qj5f5_8bi7aRk:-KOig8zzXLXySz96PssU:b-7rzchs) ([raw file](https://github.com/maidsafe/rfcs/blob/9c6abfed32d68e2bee0300090bf80a66a8254f46/text/0037-disjoint-groups/0037-disjoint-groups.md#L45)):*
<details><summary><i>Previously, maqi wrote…</i></summary>
> we better make it clear that such `group` is from each node's perspective.
> i.e. node 1 will be in node 2's G(00) but in node 3's G(01)
</details>
No, they aren't: The groups that constitute the network are absolute. Which of these groups a node or an address belongs to is absolute. If groups `G(00)` and `G(01)` both exist, a node starting with `00` will be in the former and a node starting with `01` will be in the latter. That is independent of your perspective on the network.

The only thing that differs between nodes is _which_ groups they are connected to. If I am a member of `G(011)`, I'm not connected to `G(000)`. I don't need to know about its existence and I don't need to be notified if e.g. it splits.

---

*[text/0037-disjoint-groups/0037-disjoint-groups.md, line 47 \[r1\]](https://reviewable.io:443/reviews/maidsafe/rfcs/166#-KOibADU3VVqrJ9vXzm6:-KOihdVqHI52jcRyvkgS:b-8ttob0) ([raw file](https://github.com/maidsafe/rfcs/blob/9c6abfed32d68e2bee0300090bf80a66a8254f46/text/0037-disjoint-groups/0037-disjoint-groups.md#L47)):*
<details><summary><i>Previously, maqi wrote…</i></summary>
> `not by only` -> `not only by` ?
</details>
I changed the formulation a bit. I wanted to make it clear that you need both the set of nodes and the set of groups (not just either one of them) to specify the network.

---

*[text/0037-disjoint-groups/0037-disjoint-groups.md, line 82 \[r1\]](https://reviewable.io:443/reviews/maidsafe/rfcs/166#-KOicdHz-gsTIt90Z2fn:-KOihopQPlZFHnbz_PVj:b-awqxh4) ([raw file](https://github.com/maidsafe/rfcs/blob/9c6abfed32d68e2bee0300090bf80a66a8254f46/text/0037-disjoint-groups/0037-disjoint-groups.md#L82)):*
Exactly, I didn't want to decide implementation details here, and you're right: maybe I shouldn't even have introduced the notion of a bucket here. I do think that it makes it easier to reason about the concepts, however, independently of whether it is represented in code or not (it doesn't need to be).

Basically a `HashMap<Prefix, Vec<XorName>>` would be all that's needed to represent the data: mapping prefixes to sets of group members. Which groups need to be in that map is defined by the rules specified in this RFC.

Whether in the _actual_ implementation it's more convenient to use a `Vec` and/or to sort the groups in buckets is something that shouldn't be decided in the RFC, I think.

> differ exactly in the 0th bit from 0101, i.e. all groups whose nodes' name starts with 1 ?

No, all groups whose prefix is either `1` or `11` or `110` or `1101`, or any extension of `1101`. And we need to connect to _all_ of those group's members.

Keep in mind that the "group prefix" is not something that can necessarily be derived from its members: If all nodes in a group start with `110`, that group's prefix could be `110`, but it could also be `11` or `1` or even the empty prefix. The group prefix not only defines who is a member of the group but also which data chunks it is responsible for. So it is something that needs to be made explicit and needs to be kept track of.

I'm not sure what you mean by "dynamic group filter". The criterion for which groups to connect to depends on our own current group prefix and is in that sense dynamic, yes, but I think the RFC specifies all that information: It gives the criterion, and a merge and split flow that suggests a way for how to update and communicate the group data.

---

*[text/0037-disjoint-groups/0037-disjoint-groups.md, line 135 \[r1\]](https://reviewable.io:443/reviews/maidsafe/rfcs/166#-KOieuOkLg9V0nAbvfHX:-KOiko2no-oJkjjRkF9W:b8oeb3t) ([raw file](https://github.com/maidsafe/rfcs/blob/9c6abfed32d68e2bee0300090bf80a66a8254f46/text/0037-disjoint-groups/0037-disjoint-groups.md#L135)):*
<details><summary><i>Previously, maqi wrote…</i></summary>
> Does it mean `the node needs to send a direct `GroupSplit(p)` message to all its contacts in its routing_table` ?
</details>
Yes, it does: The contacts it is connected to are _exactly_ those nodes in the network that need to be informed about this particular change in the grouping.

---

*[text/0037-disjoint-groups/0037-disjoint-groups.md, line 157 \[r1\]](https://reviewable.io:443/reviews/maidsafe/rfcs/166#-KOif7mdHdcorIqcp46u:-KOiq7Ushe71rLwYguRO:b-896fix) ([raw file](https://github.com/maidsafe/rfcs/blob/9c6abfed32d68e2bee0300090bf80a66a8254f46/text/0037-disjoint-groups/0037-disjoint-groups.md#L157)):*
<details><summary><i>Previously, maqi wrote…</i></summary>
> we better includes an example of split and merge, to illustrate how it works
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/rfcs/166)*

maqi:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[text/0037-disjoint-groups/0037-disjoint-groups.md, line 45 \[r1\]](https://reviewable.io:443/reviews/maidsafe/rfcs/166#-KOiap0qj5f5_8bi7aRk:-KOiyQykFgsH3tds0DJE:b-1ilzk0) ([raw file](https://github.com/maidsafe/rfcs/blob/9c6abfed32d68e2bee0300090bf80a66a8254f46/text/0037-disjoint-groups/0037-disjoint-groups.md#L45)):*
> No, they aren't: The groups that constitute the network are absolute. Which of these groups a node or an address belongs to is absolute. If groups `G(00)` and `G(01)` both exist, a node starting with `00` will be in the former and a node starting with `01` will be in the latter. That is independent of your perspective on the network.
> 
> The only thing that differs between nodes is _which_ groups they are connected to. If I am a member of `G(011)`, I'm not connected to `G(000)`. I don't need to know about its existence and I don't need to be notified if e.g. it splits.
</details>
well, this is fundamentally different to the current implementation.
do you mean the `buckets` in the routing_table shall be based on `absolute perspective (to all zero base address)?

---

*[text/0037-disjoint-groups/0037-disjoint-groups.md, line 82 \[r1\]](https://reviewable.io:443/reviews/maidsafe/rfcs/166#-KOicdHz-gsTIt90Z2fn:-KOidoOwT9NKp_BKcXwk:boo80k4) ([raw file](https://github.com/maidsafe/rfcs/blob/9c6abfed32d68e2bee0300090bf80a66a8254f46/text/0037-disjoint-groups/0037-disjoint-groups.md#L82)):*
> Exactly, I didn't want to decide implementation details here, and you're right: maybe I shouldn't even have introduced the notion of a bucket here. I do think that it makes it easier to reason about the concepts, however, independently of whether it is represented in code or not (it doesn't need to be).
> 
> Basically a `HashMap<Prefix, Vec<XorName>>` would be all that's needed to represent the data: mapping prefixes to sets of group members. Which groups need to be in that map is defined by the rules specified in this RFC.
> 
> Whether in the _actual_ implementation it's more convenient to use a `Vec` and/or to sort the groups in buckets is something that shouldn't be decided in the RFC, I think.
> 
> > differ exactly in the 0th bit from 0101, i.e. all groups whose nodes' name starts with 1 ?
> 
> No, all groups whose prefix is either `1` or `11` or `110` or `1101`, or any extension of `1101`. And we need to connect to _all_ of those group's members.
> 
> Keep in mind that the "group prefix" is not something that can necessarily be derived from its members: If all nodes in a group start with `110`, that group's prefix could be `110`, but it could also be `11` or `1` or even the empty prefix. The group prefix not only defines who is a member of the group but also which data chunks it is responsible for. So it is something that needs to be made explicit and needs to be kept track of.
> 
> I'm not sure what you mean by "dynamic group filter". The criterion for which groups to connect to depends on our own current group prefix and is in that sense dynamic, yes, but I think the RFC specifies all that information: It gives the criterion, and a merge and split flow that suggests a way for how to update and communicate the group data.
</details>
you are correct if the perspective is absolute, i.e. to all zero base address.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/rfcs/166)*

afck:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[text/0037-disjoint-groups/0037-disjoint-groups.md, line 45 \[r1\]](https://reviewable.io:443/reviews/maidsafe/rfcs/166#-KOiap0qj5f5_8bi7aRk:-KOj0UnGeK60jlORIpeM:b-5324mt) ([raw file](https://github.com/maidsafe/rfcs/blob/9c6abfed32d68e2bee0300090bf80a66a8254f46/text/0037-disjoint-groups/0037-disjoint-groups.md#L45)):*
<details><summary><i>Previously, maqi wrote…</i></summary>
> well, this is fundamentally different to the current implementation.
> do you mean the `buckets` in the routing_table shall be based on `absolute perspective (to all zero base address)?
</details>
No, the _buckets_ are still a relative notion: _Your_ 0-th bucket is something different than _my_ 0-th bucket.

But even in the current implementation the node names are stored in an "absolute" form (i.e. not xored with my own name), and the group prefixes are not represented at all.

Anyway, if you are `11111...`, and in group `G(11)`, and your 0-th bucket contains group `G(01)`, you will still store that as `G(01)` and not as `G(10)` (xored with your own name). So I'd say: buckets are relative, but group prefixes and node names are absolute.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/rfcs/166)*

maqi:
*[text/0037-disjoint-groups/0037-disjoint-groups.md, line 45 \[r1\]](https://reviewable.io:443/reviews/maidsafe/rfcs/166#-KOiap0qj5f5_8bi7aRk:-KOj2eQXE4YZ34xHOhpU:bry22xk) ([raw file](https://github.com/maidsafe/rfcs/blob/9c6abfed32d68e2bee0300090bf80a66a8254f46/text/0037-disjoint-groups/0037-disjoint-groups.md#L45)):*
> No, the _buckets_ are still a relative notion: _Your_ 0-th bucket is something different than _my_ 0-th bucket.
> 
> But even in the current implementation the node names are stored in an "absolute" form (i.e. not xored with my own name), and the group prefixes are not represented at all.
> 
> Anyway, if you are `11111...`, and in group `G(11)`, and your 0-th bucket contains group `G(01)`, you will still store that as `G(01)` and not as `G(10)` (xored with your own name). So I'd say: buckets are relative, but group prefixes and node names are absolute.
</details>
why we still need a `relative buckets` in routing_table? this is confusing and will complicate the calculation.
can't we make `buckets` absolute as well?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/rfcs/166)*

afck:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[text/0037-disjoint-groups/0037-disjoint-groups.md, line 45 \[r1\]](https://reviewable.io:443/reviews/maidsafe/rfcs/166#-KOiap0qj5f5_8bi7aRk:-KOj3GBArI4C98ZZJSVk:b-260895) ([raw file](https://github.com/maidsafe/rfcs/blob/9c6abfed32d68e2bee0300090bf80a66a8254f46/text/0037-disjoint-groups/0037-disjoint-groups.md#L45)):*
<details><summary><i>Previously, maqi wrote…</i></summary>
> why we still need a `relative buckets` in routing_table? this is confusing and will complicate the calculation.
> can't we make `buckets` absolute as well?
> 
</details>
Strictly speaking, the notion of a bucket is not needed at all, for anything. It might just be convenient to sort the groups we are connected to by the length of the common prefix - the "bucket index". Just like it might make sense (as an implementation detail) to store the members of the buckets sorted by distance from me.

My bucket number 3 is the set of all connected groups that share the first three bits with me, but not the fourth one. I don't know what it would mean to make that notion absolute: They are an inherently relative notion, describing how far away another group is from me.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/rfcs/166)*

maqi:
Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/rfcs/166)*

PR #170, opened 2016-08-30
Low level API

ustulation:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/170)


krishnaIndia:
*[text/0000-low-level-api/0000-low-level-api.md, line 67 \[r1\]](https://reviewable.io:443/reviews/maidsafe/rfcs/170#-KQWV7C64NU_HhSTnCzq:-KQWV7C64NU_HhSTnCzr:b-1sll4k) ([raw file](https://github.com/maidsafe/rfcs/blob/2b47b45199b3c1c1e0199433ce41c44081fddc7e/text/0000-low-level-api/0000-low-level-api.md#L67)):*
> ```Markdown
> /// **return-value**: Non-zero in case of error giving the error-code.
> #[no_mangle]
> pub unsafe extern "C" fn struct_data_create(app: *const AppHandle,
> ```

Can a function be exposed to validate ownership of the structure data?

Consider a blog like application, where the application would prefer to show additional features to the user who is owner(admin). At present, to validate the ownership, the application will have to  post the SD to determine whether the current user is the owner of the SD. If we have a utility function, it would be easy to validate by invoking that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/rfcs/170)*

ustulation:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[text/0000-low-level-api/0000-low-level-api.md, line 67 \[r1\]](https://reviewable.io:443/reviews/maidsafe/rfcs/170#-KQWV7C64NU_HhSTnCzq:-KQWYf-Sr3m5W1V-6l5D:b-896fix) ([raw file](https://github.com/maidsafe/rfcs/blob/2b47b45199b3c1c1e0199433ce41c44081fddc7e/text/0000-low-level-api/0000-low-level-api.md#L67)):*
<details><summary><i>Previously, krishnaIndia (Krishna) wrote…</i></summary>
> Can a function be exposed to validate ownership of the structure data?
> 
> Consider a blog like application, where the application would prefer to show additional features to the user who is owner(admin). At present, to validate the ownership, the application will have to  post the SD to determine whether the current user is the owner of the SD. If we have a utility function, it would be easy to validate by invoking that.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/rfcs/170)*

afck:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[text/0000-low-level-api/0000-low-level-api.md, line 333 \[r2\]](https://reviewable.io:443/reviews/maidsafe/rfcs/170#-KQZteJLbQhrMjBgL0XW:-KQZteJMIghI8o74nuz2:b-2r82w1) ([raw file](https://github.com/maidsafe/rfcs/blob/96da6fc94e572413d28cc7551522448e4dbff5ff/text/0000-low-level-api/0000-low-level-api.md#L333)):*
> ```Markdown
> pub unsafe extern "C" fn appendable_data_get_nth_data_id(app: *const AppHandle,
>                                                          ad: *const AppendableDataHandle,
>                                                          n: u64,
> ```

RFC 38 uses sets to store appended data, so there is no order apart from whatever `BTreeSet` uses internally. Should `n` refer to that?
RFC 38 also doesn't plan for a new type of `Get` response, so the whole appendable data, with all its items will be sent anyway.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/rfcs/170)*

afck:
*[text/0000-low-level-api/0000-low-level-api.md, line 333 \[r2\]](https://reviewable.io:443/reviews/maidsafe/rfcs/170#-KQZteJLbQhrMjBgL0XW:-KQa22VhB5Kq481bH_PU:bir6qvd) ([raw file](https://github.com/maidsafe/rfcs/blob/0606490dd7dee364db2198276df6af73c9954ccf/text/0000-low-level-api/0000-low-level-api.md#L333)):*
> RFC 38 uses sets to store appended data, so there is no order apart from whatever `BTreeSet` uses internally. Should `n` refer to that?
> RFC 38 also doesn't plan for a new type of `Get` response, so the whole appendable data, with all its items will be sent anyway.
</details>
As discussed, we'll go with this for now: the `n` refers to the internal order.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/rfcs/170)*

PR #174, opened 2016-08-31
merkle_tree of checkpoints for datachain

maqi:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/174)

### DIFF
--- a/text/0029-data-chains.md/0029-data-chains.md
+++ b/text/0029-data-chains.md/0029-data-chains.md
@@ -326,6 +326,33 @@ a signed checkpoint, but a checkpoint that can evaluate when all contained block
 by xoring back to all zero's. This RFC does not attempt to include any such efficient mechanism,
 instead this is left for further design improvements and RFC's.
 
+## Merkle tree of checkpoints
+
+A datachain of the group `locks` the node and data status of that group along the timeline. But this
+still leave the risk that a whole forged data chain cannot be detected by the network. A merkle-tree
+of checkpoints, which `locks` checkpoints network wide (actually creates a snapshot), can expose
+such forged chain easily. Once [DisjointGroup] is deployed, such merkle-tree can be computed as:
+
+1. Every fixed interval, each group computes a hash of all the checkpoints generated during that
+period. This hash will then be exchanged with group's sibling group to compute an one level up hash
+(say group 000 exchanging with group 001 to compute hash for 00x).
+
+2. Once one level up hash computed, the group exchange it with a group it knows in that level, so
+the hash of that level can be computed (exchange hash of 00x with a group in 01 to compute 0x hash).
+
+3. Repeat this procedure till get a root hash.
+
+4. To verify the computed merkle-tree is correct, the root hash needs to be published to the network
+(or just contact the furthest group knows, which has the highest chance of mismatch, to reduce the
+messages transferred).
+
+Such merkle-trees (snapshots of network) prove the validity of datachains when restoring data
+during network restart / mass segmentation. To prove identity to the network, a member of group 000
+only needs to provide: root hash, hash of 0x and 1x, hash of 00x and 01x, checkpoints hashes. If the
+duty of providing this set of hashes left to the node joining the network, the existing nodes only
+need to holds the root hash in cache (and probably store the partial merkel-tree containing itself
+in disk, so the joining node can get access to via fetching request).
+
 # Drawbacks
 
 - In very small networks (less than approx 3000) network difficulty is a fluctuating number, this can
@@ -354,3 +381,4 @@ Not initially required, but should be considered in near future.
 [Proof]: https://dirvine.github.io/data_chain/master/data_chain/chain/node_block/struct.Proof.html
 [BlockIdentifier]: https://dirvine.github.io/data_chain/master/data_chain/enum.BlockIdentifier.html
 [create_link_descriptor()]: https://dirvine.github.io/data_chain/master/data_chain/chain/node_block/fn.create_link_descriptor.html
+[DisjointGroup]: https://github.com/maidsafe/rfcs/blob/master/text/0037-disjoint-groups/0037-disjoint-groups.md


afck:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[text/0029-data-chains.md/0029-data-chains.md, line 334 \[r1\]](https://reviewable.io:443/reviews/maidsafe/rfcs/174#-KQWRL2CZLqoh8tjUFX7:-KQWRL2CZLqoh8tjUFX8:b7yypd2) ([raw file](https://github.com/maidsafe/rfcs/blob/190b17d44ff367eee0c870097545bff43b5d5948/text/0029-data-chains.md/0029-data-chains.md#L334)):*
> ```Markdown
> 
> 3. Once all hashes from all groups received, each group calculates the merkle-tree, then broadcast
> the root hash to the network
> ```

But we don't need the hashes from _all_ groups to compute the Merkle tree's root hash! That would be an amount of data growing linearly with the network size, where `O(log(n))` should be achievable:

E.g. nodes in group `000` would:
* exchange their latest link's hash with `001` and vice-versa, and compute the hash of those two hashes - called hash `00` -,
* exchange hash `00` with their bucket-1- contacts (probably `010`), who share _their_ `01`-hash (probably computed from `010` and `011`, but we don't need to know that) with us, so we can both compute the `0`-hash,
* exchange the `0`-hash with our bucket-0-contacts and get the `1`-hash and compute the current network's root hash.
Then every node in the network has the root hash and all the partial hashes that lead to their latest link. So every node has an `O(log n)`-proof for its identity that every other node in the network can verify.

Specifically, if I (a member of `000`) want to prove my identity to any other node in the network that has the root hash, I just need to provide the `0`- and `1`-hash, so it can verify that the root is the hash of those two. Then I provide the `00`- and `01`-hashes, so it can verify that the `0`-hash is the hash of those two. Then I provide the `001`- and `000`-hashes, and finally I provide my link block (i.e. the list of my group members), so the other node can verify that the `000`-hash is actually the hash of that group.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/rfcs/174)*

afck:
Review status: 0 of 1 files reviewed at latest revision, 4 unresolved discussions.

---

*[text/0029-data-chains.md/0029-data-chains.md, line 332 \[r2\]](https://reviewable.io:443/reviews/maidsafe/rfcs/174#-KQdg4kZSAacGtLwuElh:-KQdg4kZSAacGtLwuEli:b-icro46) ([raw file](https://github.com/maidsafe/rfcs/blob/23ab761f15894c3d4e91600884050407485287d4/text/0029-data-chains.md/0029-data-chains.md#L332)):*
> ```Markdown
> 
> A datachain of the group `locks` the node and data status of that group along the timeline. But this
> still leave the risk that a whole forged blockchain cannot be detected by the network. A merkle-tree
> ```

"blockchain" -> "data chain"?

---

*[text/0029-data-chains.md/0029-data-chains.md, line 336 \[r2\]](https://reviewable.io:443/reviews/maidsafe/rfcs/174#-KQdgViFy_GI9FlDq_sY:-KQdgViGAaS_cJWeDawX:b-m70pla) ([raw file](https://github.com/maidsafe/rfcs/blob/23ab761f15894c3d4e91600884050407485287d4/text/0029-data-chains.md/0029-data-chains.md#L336)):*
> ```Markdown
> such forged chain easily. Once [DisjointGroup] is deployed, such merkle-tree can be computed as:
> 
> 1. Every fixed interval (say one hour), each group computes a hash of all the checkpoints generated
> ```

I don't think it even needs to take all the checkpoints. Wouldn't the latest link block suffice?

---

*[text/0029-data-chains.md/0029-data-chains.md, line 347 \[r2\]](https://reviewable.io:443/reviews/maidsafe/rfcs/174#-KQdgpepEthaPmLIQivc:-KQdgpeqBwiHfA9t22Gu:b4fzxaq) ([raw file](https://github.com/maidsafe/rfcs/blob/23ab761f15894c3d4e91600884050407485287d4/text/0029-data-chains.md/0029-data-chains.md#L347)):*
> ```Markdown
> 4. To verify the computed merkle-tree is correct, the root hash needs to be published to the network
> (or just contact the furthest group knows, which has the highest chance of mismatch, to reduce the
> messages transferred).
> ```

So this is just a check that nothing got out of sync, is it? Because every group should already have computed the same root hash by itself?

---

*[text/0029-data-chains.md/0029-data-chains.md, line 349 \[r2\]](https://reviewable.io:443/reviews/maidsafe/rfcs/174#-KQdh4pFD1rmuvt2vUH7:-KQdh4pFD1rmuvt2vUH8:b-ac2twt) ([raw file](https://github.com/maidsafe/rfcs/blob/23ab761f15894c3d4e91600884050407485287d4/text/0029-data-chains.md/0029-data-chains.md#L349)):*
> ```Markdown
> messages transferred).
> 
> Such merkle-trees (snapshots of network) prove the validatity of datachains when restoring data
> ```

"validity"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/rfcs/174)*

maqi:
Review status: 0 of 1 files reviewed at latest revision, 4 unresolved discussions.

---

*[text/0029-data-chains.md/0029-data-chains.md, line 336 \[r2\]](https://reviewable.io:443/reviews/maidsafe/rfcs/174#-KQdgViFy_GI9FlDq_sY:-KQez4tYG6c2syLRpS8E:b4ugb9m) ([raw file](https://github.com/maidsafe/rfcs/blob/23ab761f15894c3d4e91600884050407485287d4/text/0029-data-chains.md/0029-data-chains.md#L336)):*
> I don't think it even needs to take all the checkpoints. Wouldn't the latest link block suffice?
</details>
not sure here. as I understand, link covers some checkpoints. these may cause two issues:
1, there might be multiple links during an interval, so only `latest` might not be enough.
2, the link might be too long, which causing a wider range of data lost in case of a hash mismatch

any way, we can easily define anything to be `locked` by the merkle tree later on, so I will prefer just staying with the current wording.

---

*[text/0029-data-chains.md/0029-data-chains.md, line 347 \[r2\]](https://reviewable.io:443/reviews/maidsafe/rfcs/174#-KQdgpepEthaPmLIQivc:-KQf-Jio4y2boZJjmXcr:bb7sa0c) ([raw file](https://github.com/maidsafe/rfcs/blob/23ab761f15894c3d4e91600884050407485287d4/text/0029-data-chains.md/0029-data-chains.md#L347)):*
> So this is just a check that nothing got out of sync, is it? Because every group should already have computed the same root hash by itself?
</details>
yes.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/rfcs/174)*

afck:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[text/0029-data-chains.md/0029-data-chains.md, line 336 \[r2\]](https://reviewable.io:443/reviews/maidsafe/rfcs/174#-KQdgViFy_GI9FlDq_sY:-KQf0HmfNnJ1_6_qy_EP:b-dw3a5s) ([raw file](https://github.com/maidsafe/rfcs/blob/23ab761f15894c3d4e91600884050407485287d4/text/0029-data-chains.md/0029-data-chains.md#L336)):*
<details><summary><i>Previously, maqi wrote…</i></summary>
> not sure here. as I understand, link covers some checkpoints. these may cause two issues:
> 1, there might be multiple links during an interval, so only `latest` might not be enough.
> 2, the link might be too long, which causing a wider range of data lost in case of a hash mismatch
> 
> any way, we can easily define anything to be `locked` by the merkle tree later on, so I will prefer just staying with the current wording.
</details>
OK. Not sure I understand what exactly "checkpoint" means in this context.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/rfcs/174)*

dirvine:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[text/0029-data-chains.md/0029-data-chains.md, line 336 \[r3\]](https://reviewable.io:443/reviews/maidsafe/rfcs/174#-KQf1s7XftN2DwD43VJL:-KQf1s7XftN2DwD43VJM:b-newazb) ([raw file](https://github.com/maidsafe/rfcs/blob/3ecaa6985c8efe0792d0c6949c6cf92c529d3d28/text/0029-data-chains.md/0029-data-chains.md#L336)):*
> ```Markdown
> such forged chain easily. Once [DisjointGroup] is deployed, such merkle-tree can be computed as:
> 
> 1. Every fixed interval (say one hour), each group computes a hash of all the checkpoints generated
> ```

I would add this interval should be event based and not time based. Then we should be good to merge this further work section.  IF we decided to go this route this would be a separate RFC so maybe worth mentioning that as well?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/rfcs/174)*

maqi:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[text/0029-data-chains.md/0029-data-chains.md, line 336 \[r2\]](https://reviewable.io:443/reviews/maidsafe/rfcs/174#-KQdgViFy_GI9FlDq_sY:-KQf29hWazkKsXD9XbfM:b-gtoint) ([raw file](https://github.com/maidsafe/rfcs/blob/23ab761f15894c3d4e91600884050407485287d4/text/0029-data-chains.md/0029-data-chains.md#L336)):*
> OK. Not sure I understand what exactly "checkpoint" means in this context.
</details>
to me, this checkpoint means a locker needs to be locked not only along the time (by the data chain) but also cross locked with other chain. It can be link, data or anything (combined or just one).

---

*[text/0029-data-chains.md/0029-data-chains.md, line 336 \[r3\]](https://reviewable.io:443/reviews/maidsafe/rfcs/174#-KQf1s7XftN2DwD43VJL:-KQf2quYr2fjw_eyLkWl:bbzt8nm) ([raw file](https://github.com/maidsafe/rfcs/blob/3ecaa6985c8efe0792d0c6949c6cf92c529d3d28/text/0029-data-chains.md/0029-data-chains.md#L336)):*
<details><summary><i>Previously, dirvine (David Irvine) wrote…</i></summary>
> I would add this interval should be event based and not time based. Then we should be good to merge this further work section.  IF we decided to go this route this would be a separate RFC so maybe worth mentioning that as well?
</details>
the problem of event based is a synchronization among data chains. as each group having there own event sequence.
Also, when trying to validate a checkpoint, event based requires an explicit merkel tree snapshot index to be recorded for each checkpoint, meanwhile the interval based can be deduced.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/rfcs/174)*

dirvine:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[text/0029-data-chains.md/0029-data-chains.md, line 336 \[r3\]](https://reviewable.io:443/reviews/maidsafe/rfcs/174#-KQf1s7XftN2DwD43VJL:-KQf6gkB4CmnvRmXGssd:b-e2hhm5) ([raw file](https://github.com/maidsafe/rfcs/blob/3ecaa6985c8efe0792d0c6949c6cf92c529d3d28/text/0029-data-chains.md/0029-data-chains.md#L336)):*
<details><summary><i>Previously, maqi wrote…</i></summary>
> the problem of event based is a synchronization among data chains. as each group having there own event sequence.
> Also, when trying to validate a checkpoint, event based requires an explicit merkel tree snapshot index to be recorded for each checkpoint, meanwhile the interval based can be deduced.
</details>
Yes this is why we would present a full RFC for this further work IMHO,  time is always simple but IMO rarely correct in event driven systems. Checkpoints should not change frequently unless we had split/merge frequency that was high.  To me that would be a design error.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/rfcs/174)*

maqi:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[text/0029-data-chains.md/0029-data-chains.md, line 336 \[r3\]](https://reviewable.io:443/reviews/maidsafe/rfcs/174#-KQf1s7XftN2DwD43VJL:-KQf8LaiqKhT4r8H2I7L:bcpruuk) ([raw file](https://github.com/maidsafe/rfcs/blob/3ecaa6985c8efe0792d0c6949c6cf92c529d3d28/text/0029-data-chains.md/0029-data-chains.md#L336)):*
<details><summary><i>Previously, dirvine (David Irvine) wrote…</i></summary>
> Yes this is why we would present a full RFC for this further work IMHO,  time is always simple but IMO rarely correct in event driven systems. Checkpoints should not change frequently unless we had split/merge frequency that was high.  To me that would be a design error.
</details>
I agree a full RFC shall be raised to detailed this further work. However, I think here a vague word of `interval` shall be OK to present the concept. `time interval` is just used as a simple example, which I think shall be fine here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/rfcs/174)*

dirvine:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[text/0029-data-chains.md/0029-data-chains.md, line 336 \[r3\]](https://reviewable.io:443/reviews/maidsafe/rfcs/174#-KQf1s7XftN2DwD43VJL:-KQfDvlZ7iMd3OkryHuk:b-jsnjll) ([raw file](https://github.com/maidsafe/rfcs/blob/3ecaa6985c8efe0792d0c6949c6cf92c529d3d28/text/0029-data-chains.md/0029-data-chains.md#L336)):*
<details><summary><i>Previously, maqi wrote…</i></summary>
> I agree a full RFC shall be raised to detailed this further work. However, I think here a vague word of `interval` shall be OK to present the concept. `time interval` is just used as a simple example, which I think shall be fine here?
</details>
Time further specifies the interval type and right now I don't think we should though.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/rfcs/174)*

dirvine:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[text/0029-data-chains.md/0029-data-chains.md, line 334 \[r1\]](https://reviewable.io:443/reviews/maidsafe/rfcs/174#-KQWRL2CZLqoh8tjUFX7:-KQfE1oCz_VEletGm3DU:b-67r7p5) ([raw file](https://github.com/maidsafe/rfcs/blob/190b17d44ff367eee0c870097545bff43b5d5948/text/0029-data-chains.md/0029-data-chains.md#L334)):*
> But we don't need the hashes from _all_ groups to compute the Merkle tree's root hash! That would be an amount of data growing linearly with the network size, where `O(log(n))` should be achievable:
> 
> E.g. nodes in group `000` would:
> * exchange their latest link's hash with `001` and vice-versa, and compute the hash of those two hashes - called hash `00` -,
> * exchange hash `00` with their bucket-1- contacts (probably `010`), who share _their_ `01`-hash (probably computed from `010` and `011`, but we don't need to know that) with us, so we can both compute the `0`-hash,
> * exchange the `0`-hash with our bucket-0-contacts and get the `1`-hash and compute the current network's root hash.
> Then every node in the network has the root hash and all the partial hashes that lead to their latest link. So every node has an `O(log n)`-proof for its identity that every other node in the network can verify.
> 
> Specifically, if I (a member of `000`) want to prove my identity to any other node in the network that has the root hash, I just need to provide the `0`- and `1`-hash, so it can verify that the root is the hash of those two. Then I provide the `00`- and `01`-hashes, so it can verify that the `0`-hash is the hash of those two. Then I provide the `001`- and `000`-hashes, and finally I provide my link block (i.e. the list of my group members), so the other node can verify that the `000`-hash is actually the hash of that group.
</details>
@maqi  ^^

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/rfcs/174)*

maqi:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[text/0029-data-chains.md/0029-data-chains.md, line 334 \[r1\]](https://reviewable.io:443/reviews/maidsafe/rfcs/174#-KQWRL2CZLqoh8tjUFX7:-KQfIBu_4QekeuZLXHRB:bip2my8) ([raw file](https://github.com/maidsafe/rfcs/blob/190b17d44ff367eee0c870097545bff43b5d5948/text/0029-data-chains.md/0029-data-chains.md#L334)):*
<details><summary><i>Previously, dirvine (David Irvine) wrote…</i></summary>
> @maqi  ^^
</details>
yes, the doc has been updated according to this comment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/rfcs/174)*

dirvine:
Reviewed 1 of 1 files at r3.
Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/rfcs/174)*

dirvine:
Reviewed 1 of 1 files at r4.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/rfcs/174)*

PR #177, opened 2016-09-01
locked & compressed data chains

dirvine:
A proposal to enable chains to lock in place and remove block proofs on churn event agreement. Will require some testing and perhaps team discussion on accumulating agreement on blocks, but this is not a huge issue, more one of implementation options.

---

This change is [Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/177)

### DIFF
--- a/0041-locked-data-chains/0041-locked-data-chains.md
+++ b/0041-locked-data-chains/0041-locked-data-chains.md
@@ -1,0 +1,54 @@
+# Locked DataChains
+
+- Status: proposed 
+- Type: enhancement
+- Related components: data_chain
+- Start Date: 01-09-2016
+- Discussion: (fill me in with link to RFC discussion - shepherd will complete this) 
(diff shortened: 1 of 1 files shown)


frabrunelle:
Number 41 is used by the [Low Level API RFC](https://github.com/maidsafe/rfcs/blob/master/text/0041-low-level-api/0041-low-level-api.md). So this RFC would need a new number 😃

dirvine:
Ah no worries @frabrunelle  I am happy if you wish to move this to 0043 (I have another one for 42 I need to commit this weekend)

frabrunelle:
42 and 43 are also used 😛 So you could use 44 and 45 😄

dirvine:
Ah I will do thanks @frabrunelle

dhardy:
Swap the "Motivation" and "Summary" titles, maybe?

PR #189, opened 2016-09-20
Update to proposed v0.6 launcher api

krishnaIndia:
- Mapping to one to one functions of safe_core
- Adding DataIdentifier endpoint

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/rfcs/189)


text/0042-launcher-api-v0.6/api/appendable_data.md
```diff
@@ -39,39 +37,71 @@ Authorization: Bearer <TOKEN>
...
+
+#### Body
+```
+{
+  handleId:u64 number representing the appendable data handle id
+}
+```
+
+## Save AppendableData
+
```
frabrunelle: I think we should add a text similar to what you added in Structured Data. something like this:

> The safe_core calls PUT to store the data in the network and POST to update an existing data in the network. When an appendable data is created for the first time, the application should call the PUT endpoint, and while updating an appendable data, the application should call the POST endpoint.

PR #198, opened 2016-10-31
MAID-1621 use a more efficient message routing algorithm

afck:


text/0037-disjoint-groups/0037-disjoint-groups.md
```diff
@@ -91,36 +91,24 @@ To avoid repeated splitting and merging due to small fluctuations in group size,
 
 ### Message routing
 
-To relay a message from an individual node on a given `route` in a node `n` in group `G(p)` for a destination _node_ `d`:
+When learning about a change in any group (new member, group split or merge), each node signs the sorted new list of members in _that_ group and sends it to each member of _its own_ group, so every node has a signature from each of its group peers, for the list of current members of each neighbouring group.
```
maqi: not sure but i think the last `each neighbouring group` better to be changed to `that changing neighbouring group`.

the knowledge of `each neighbouring group is the final result, but I think here we are just talking about the flow of updating just one group?

afck: Good point. I named the groups `A` and `B`. It's not exactly poetry now, but maybe less confusing. :grin:

text/0037-disjoint-groups/0037-disjoint-groups.md
```diff
@@ -91,36 +91,24 @@ To avoid repeated splitting and merging due to small fluctuations in group size,
...
 
-Assume a node `n` in group `B` receives a group message signed by someone in group `A`:
+* a quorum of the entries of `hop_signatures[0]`, if the source authority is a group, or
+* one entry of `hop_signatures[0]`, corresponding to the source, if the source authority is a single node.
 
-* If `A` is not connected to `B`, drop the message.
-* If this message has not yet been signed by a quorum of members of `A`, keep it in the cache.
-* If it has been signed by a quorum of members of `A`: If we are the recipient, handle it. If not, find the connected group `C` closest to the destination and send the message, signed by `n`, to _every member_ of `C`.
+If a group sends a message, only the `route`-th node sends the full `HopMessage`, and the others only a `HopMessageSignature`. The recipient keeps the full message in cache until it has collected a quorum of signatures. Then it continues relaying the message as detailed below. If an individual node sends a message, it sends the full `HopMessage` with only its own signature.
```
dirvine: Makes sense to me, thanks @afck

PR #199, opened 2016-11-02
unversioned = 500, versioned = 501

frabrunelle:
See https://github.com/maidsafe/safe_core/blob/master/src/lib.rs#L73-L76

It would be good to fix this just to avoid any confusion 🙂


frabrunelle:
@krishnaIndia are you OK with merging this?

PR #200, opened 2016-11-04
Sub-group RFC

dhardy:
[Rendered](https://github.com/dhardy/rfcs-1/blob/master/text/0000-subgroups.md)


dhardy:
Noted caveat: this assumes groups may be much larger than the minimum group size (which is currently the case anyway). This RFC does not address how to make routing work between many nodes without requiring `O(n^2)` connections.

PR #209, opened 2016-11-23
Updates to RFC 46: Structs-based Auth Flow & updated internal Info

gnunicorn:
Based on the work of @krishnaIndia .

### DIFF
--- a/text/0046-new-auth-flow/details.md
+++ b/text/0046-new-auth-flow/details.md
@@ -6,7 +6,7 @@ Outlining the specific network actions, activities and things the authenticator 
 
  - On successful registration the user `RootContainer` and `AccessContainer`  must be created at random locations
  - A symmetric key must be generated for encrypting those containers
- - The dataId of each, together with the must be stored in the Account on the network.
+ - The DataIdentifier of each, together with the must be stored in the Account on the network.
  - The root container will start out with this default set of keys, for each a random location must be chosen and a set of encryption keys generated:
 
 ```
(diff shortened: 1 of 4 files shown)


text/0046-new-auth-flow/details.md
```diff
@@ -0,0 +1,74 @@
...
+
+
+## App Authorisation flow
+
+App will send the authorisation request to the authenticator to the authorisation endpoint, `safe-auth://action:payload?riq={random-id}` as defined in the [authentication-protocol appendix](./authentication-protocol.md). Upon receiving such a request, the authenticator may grant directly access if the user has granted this app access before, or must prompt the user to grant access. Should the user grant access the authenticator must:
+
+ 1. generate a random sign key pair for the app
+ 2. generate a random encryption key
+ 3. register the public sign key with MaidManager
+ 4. If the app is requesting for any container access, the public sign key must added to the container along with the requested permissions
```
frabrunelle: to which container?

frabrunelle:
why is the `version` field being removed?

gnunicorn:
@frabrunelle it has only cosmetic purposes at the moment: to show a nice message to the user. However, @krishnaIndia pointed out correctly, that because we don't require any of the app information when doing progressive container auth, we might show outdated info - saying it is version `0.0.1` while in fact the app was upgraded in between and now runs as `0.2.0` or something. While it is unlikely that app name and vendor change, this is very likely to happen. So if the information isn't accurate, what purpose does it serve? @krishnaIndia preferred to remove it altogether as it does more harm than it helps,

PR #217, opened 2018-02-20
Fix links

iancoleman:
The directory structure was modified but some links still use urls with the old directory structure (and are broken links because of that).


david-maidsafe:
@iancoleman - I'm just starting to pick up all these issues.
Thanks for doing this.

PR #221, opened 2018-06-20
MAID-2859: fix minor copy-paste errors and clarify meta voting process

Fraser999:


text/0049-parsec/0049-parsec.md
```diff
@@ -1053,7 +1054,7 @@ estimates are promoted to b's
 bin_values.
 
 Because both binary values entered
-bin_values in the same gossip event,
```
pierrechevalier83: :relieved:

text/0049-parsec/0049-parsec.md
```diff
@@ -294,12 +295,12 @@ The set of estimates of a `GossipEvent` is the set of estimates of its `self_par
 - the step number is different from the `self_parent`'s step number, in which case the estimate is updated according to the rules defined in the "concrete coin protocol"
 
 The set `bin_values` of a `GossipEvent` is the set `bin_values` of its `self_parent`, except if
-- this `GossipEvent`'s decided value is not `None`, in which case the set of estimate is the set containing only that value
+- this `GossipEvent`'s decided value is not `None`, in which case the set `bin_values` is the set containing only that value
```
pierrechevalier83: Oh oh! Copy-paste error! Good catch!

text/0049-parsec/0049-parsec.md
```diff
@@ -187,6 +187,7 @@ We need to ensure that all nodes decide upon the same order of stable `Block`s.
 
 To make a decision, we turn this single problem into a number of binary Byzantine problems. Each binary problem is effectively asking "Should this voter's opinion be considered when choosing the next stable block?". To be more specific, for each current valid voter, we ask the question "Can our latest sync event strongly see any vote for a valid, not-yet-stable block by this voter?". The answer to that question is a virtual binary value that we define as our meta vote for this voter's right to participate in the decision.
 - Note: by "virtual", we mean that there is no actual explicit "meta vote" initiated by any node. These meta votes are implicit properties of gossip. The meaning of "meta vote" is assigned to a `GossipEvent` in retrospect by a node interpreting the gossip graph. This definition of "virtual" also applies to estimates, auxiliary values, `bin_values` and decided values.
+- Note also that in the following description of the details of meta voting, a single `GossipEvent` may have multiple steps for the same meta vote associated with it.
```
pierrechevalier83: I have an issue with using "step" here as it conflicts with the concept of rounds and steps, and I don't think you're using it to mean that.

Fraser999: Well, I did really mean step, although I'd have liked to say step and round, but thought that was more confusing since we haven't defined step and round at this point.  I guess what's missing from the RFC is the definition of a something like the struct we have coded called `MetaVote`, since in the RFC a meta vote is more like the entire election for single binary value, right?  Any suggestions on how this could be better phrased?

pierrechevalier83: Lemme give it a try:

"Note also that in the following description of the details of meta voting, the values: "estimates, bin_values and auxiliary values" are each associated to a round and step. Each `GossipEvent` may carry multiple sets of each of these values (one for each step/round combination) until a decision is reached. The decided values apply for any subsequent step and round."

Now in these words, it may be better to write it a bit later in the text when things have been defined a bit more (?)

Fraser999: Yup, looks good.

PR #275, opened 2018-09-17
Safecoin Alternative Implementation

Fraser999:


text/0000-safecoin-alternative-implementation/0000-safecoin-alternative-implementation.md
```diff
@@ -0,0 +1,238 @@
...
+```rust
+struct CoinAccount {
+    id: PublicSigningKey,
+    balance: Coin,
+}
+```
+
+As with other network addressable entities, the `id` will represent the account's address on the network.  It will be managed by the close group of Vaults for that address, using a new persona: CoinManager.  The ID is also an asymmetric public signing key.
+
+A single client can create many `CoinAccount`s if desired, and there is no need to associate any of these with its main MAID account.  For example, a user may want to run multiple farmer Vaults and allocate a different `CoinAccount` for each to receive the farming rewards.
```
dirvine: Possibly we should define an account in more detail, but also what a MAID is?

text/0000-safecoin-alternative-implementation/0000-safecoin-alternative-implementation.md
```diff
@@ -0,0 +1,238 @@
...
+A CoinManager receiving a `GetTransaction` will always return a `Transaction`.  If the specified `CoinAccount` doesn't exist, `NoSuchCoinAccount` is returned.  If the `CoinAccount` does exist, but the requested `transaction_id` doesn't exist in its FIFO queue, `NoSuchTransaction` is returned.  Otherwise the appropriate `Success` value is returned.  As this message is signed by the Vaults comprising the CoinManagers, a `Transaction::Success` can be used as proof by any client, (in particular the sender) that the transaction was successful.
+
+This has the benefit of requiring neither the sender nor the receiver to stay connected while the payment is being processed, as would be the case for example if the sender needed to receive a response in order to prove payment had been made.  It also serves to further anonymise the sender, as the recipient will likely not be aware of the actual `CoinAccount` used by the sender to credit its own; only the `transaction_id` is visible.
+
+
+
+## Drawbacks
+
+A potential drawback of this proposal is that coins which are recycled aren't replaced back to the section where they were initially farmed.
+
```
dirvine: I see that a feature and a good one :-)

text/0000-safecoin-alternative-implementation/0000-safecoin-alternative-implementation.md
```diff
@@ -0,0 +1,238 @@
...
+The actual algorithm defining the farming rate is somewhat orthogonal to this RFC and worthy of its own RFC, so the details will be omitted, but it would seem fairly critical that it ensures sections don't become fully farmed or fully unfarmed.
+
+One benefit of providing coin divisibility is that on a successful farming attempt, it is trivial to pay all farmers in that section rather than the lottery approach taken in the alternative proposal whereby only one farmer is paid at a time, and only if the chosen coin is unfarmed.  Indeed, they could be paid proportional to the value they have provided to the network (this could be measured initially by node age at the moment of payment).  They could also be paid smaller amounts more frequently to encourage new Vault owners to continue to run their Vaults.
+
```
dirvine: Would this then cause issues with nodes that hold more data than others or vice versa? I mean would this code out that possibility?

text/0000-safecoin-alternative-implementation/0000-safecoin-alternative-implementation.md
```diff
@@ -0,0 +1,238 @@
...
+```
+
+The details of the `PaymentAccount` will be explained below at "[Paying for Network Resources](#paying-for-network-resources)".  Until we reach that point, just consider this as a wrapped `CoinAccount` and when `CoinAccount` is mentioned it means either of these types.
+
+
+
+### Farming
+
+Each section of the network will be responsible for a proportion of the total issuable safecoin equal to the proportion of the network address space it manages.  For instance, at the start when the network has a single section covering the full address space, that section will be responsible for all `2^32` safecoin.  When the section splits, each half will be responsible for `2^31` safecoin.
+
```
dirvine: Hmm, I am not 100% this is required to be this way. A section can say a farm attempt is valid, then query the section holding the coins of that address range to find if there is a safecoin available at the address of the farm attempt trigger? 
i.e. far attempt is hash of stuff == some value 
If successful we ask the group near `some value` if there is a safecoin of this `some value` available, or has it been farmed. This is again the lottery approach, but not sure it is still not valid in this proposal?

text/0000-safecoin-alternative-implementation/0000-safecoin-alternative-implementation.md
```diff
@@ -0,0 +1,238 @@
...
+This means that for a section with Prefix length `n`, it will be responsible for `2^(32-n)` safecoin.
+
```
dirvine: We probably need to define the responsibility here. i.e. responsible for 
Managing the issuance of coins 
Managing recycling of coins
etc.

text/0000-safecoin-alternative-implementation/0000-safecoin-alternative-implementation.md
```diff
@@ -0,0 +1,238 @@
...
+Each section's CoinManagers will maintain a record `farmed` (of type `Coin`) of the amount of safecoin farmed at that section.  Any changes to this total will come as a result of an action having passed through Parsec in order to ensure that all managers maintain an eventually consistent record.
+
+The section's `farmed` value will never be allowed to exceed the amount of coins for which that section is responsible.  Ideally no section should ever get close to "running out" of farmable coins; getting the farming rate algorithm correct should ensure that.  However, in the case that a section has farmed all of its coins, it will stop issuing any more until the `farmed` value reduces again.
+
+The actual algorithm defining the farming rate is somewhat orthogonal to this RFC and worthy of its own RFC, so the details will be omitted, but it would seem fairly critical that it ensures sections don't become fully farmed or fully unfarmed.
+
```
dirvine: possibly worth a simplified version here that states 
As resources drop below current data size + 50% or something we increase FR and at over data size + 50% we decrease farming rate.

text/0000-safecoin-alternative-implementation/0000-safecoin-alternative-implementation.md
Fraser999: Hmm, yeah - that's a tricky one to nail down.  I suppose I really mean "responsible for managing the amount of farmed coin from that address space by ensuring it doesn't exceed the allotted amount or drop below zero."  I think I could fit that in as the second sentence of the first paragraph of this section, if that makes sense?

text/0000-safecoin-alternative-implementation/0000-safecoin-alternative-implementation.md
Fraser999: I was hoping to avoid any details here since I think it'd be a bit of a distraction from the main focus, and this RFC doesn't tie us to any particular algorithm.  What I really wanted to get across was that the chosen algorithm should ensure as far as possible that we don't run out of farmable coins or farmed coins.

text/0000-safecoin-alternative-implementation/0000-safecoin-alternative-implementation.md
Fraser999: I don't think so.  In fact, because of the coin divisibility and how inexpensive it will be to allocate a farmed amount, I think this gives us more flexibilty to choose a farming rate/farming attempt algorithm which is fairest to all nodes.

As with skipping details of the farming attempt algorithm above, I should probably remove the specific example here mentioning node age, since that's a distraction which doesn't help make the intended point.  I could just as well have mentioned some other metric like data stored/number of successful Get requests/etc.

text/0000-safecoin-alternative-implementation/0000-safecoin-alternative-implementation.md
Fraser999: Yeah, I'm not convinced this is a bad thing either, but it was a point Qi made when we were discussing my approach recently 😄  I think you're right; I'll probably remove this and let Qi make the case himself.

text/0000-safecoin-alternative-implementation/0000-safecoin-alternative-implementation.md
dirvine: I understand that desire and tend to agree with it, but this needs debating in some more detail, perhaps.

text/0000-safecoin-alternative-implementation/0000-safecoin-alternative-implementation.md
dirvine: Yip, sounds good to me.

text/0000-safecoin-alternative-implementation/0000-safecoin-alternative-implementation.md
```diff
@@ -0,0 +1,238 @@
...
+(The `transaction_id` will be explained in the "[Transferring to a Different `CoinAccount`](#transferring-to-a-different-coinaccount)" section below, but when sending farming rewards, it will always be `None`.)
+
+The CoinManager persona will be responsible for managing all aspects of farming within their section.  This will include among other things:
+
+* calculating the farming rate
+* maintaining a mapping of Vault to `CoinAccount::id` for each Vault in their section (i.e. where to pay successful farming attempts)
+* sending `Credit`s to these `CoinAccount`s when a farming attempt is successful, and increasing the `farmed` total for the section
+* receiving `Credit`s from other sections and updating the corresponding `CoinAccount`s
+
+When a section `X` splits into `X0` and `X1`, each new section will start with half of `X`'s final `farmed` value.  If `farmed.parts` is odd, rather than rounding, `X1` will be allocated the extra 1 part.
```
iancoleman: Does this need to iterate over all CoinAccounts and work out the balance for each new section? What if the `X0` accounts have 55% of the farmed coins in the section? Halving the `farmed` total wouldn't be appropriate in that case, wouldn't it need to be a 55:45 split?

Fraser999: No, the `farmed` value is maintained separately to the `CoinAccount`s a section manages.  If `X` has a pre-split `farmed` value of say 100 safecoin exactly, then `X0` and `X1` will start with `farmed` values of 50 safecoin each, regardless of how many `CoinAccount`s each new section ends up managing.

PR #281, opened 2018-11-13
XOR-URLs: resolvable XOR-name based URLs

bochaco:
Proposal for having a resolver function in the SAFE app client API that allows us to have standardised `safe://`-URLs which are generated based on the XOR address of the content being referenced.

For an overview of the proposal, and foresight about its potential, [this screencast video](https://www.youtube.com/watch?v=BikfxRNARnM) explores some ideas around it with a working proof of concept.


text/0000-xor-urls/0000-xor-urls.md
```diff
@@ -0,0 +1,115 @@
...
+
+### ImmutableData XOR-URL
+XOR-URLs for ImmutableData’s are the simplest ones since they don’t need any additional information to uniquely identify them on the network, as opposed to MutableData’s that also have a type tag. Therefore an ImmutableData XOR-URL can be simply defined as `safe://<encoded ImmutableData XOR addr>`.
+
+### MutableData XOR-URL
+As already mentioned above, a string based on the XOR address along with a type tag is needed to uniquely identify a MutableData on the network, therefore the XOR-URL for a MD needs to include the type tag in it.
+
+When a MutableData is fetched, if it’s not an NFS/Files container with an `index.html` file, the resolver (e.g. the `webFetch` function) can return the MutableData’s raw data so the browser (or any client application) can render it in a specific way (see below for more details of what’s being proposed here in this regard).
+
+### MutableData versions
```
joshuef: I would make these `####` sub headers of `Mutable Data XOR-URL`

text/0000-xor-urls/0000-xor-urls.md
```diff
@@ -0,0 +1,115 @@
...
+
+The version value can be used to enforce a specific version to be retrieved, and otherwise fail if that specific version is not found. On the other hand, the latest version will be retrieved if the version value is omitted from the XOR-URL.
+
+### Paths
+Given that a referenced MutableData can effectively be the root NFS/Files container of the files of a safesite, any MD XOR-URL can also specify a path which needs to be resolved, just like when using the DNS/public-names system to resolve the path in a URL.
+
+### Browsable content
+Currently, when a public-name URL is resolved to an NFS/Files container which doesn’t have an `index.html` file, the browser simply shows an error stating that the safesite content was not found.
+
+For a MD XOR-URL that doesn’t have a path and there is no `index.html` file, the resolver can return the raw content of the MD (i.e. its key-value entries), and the browser can automatically generate an HTML page which makes the content browsable, generating links to other data when an entry’s key/value is a `safe://` string, in an analogous/similar way to how web servers on the current internet allow browsing on folders/directories.
```
joshuef: > For a MD XOR-URL that doesn’t have a path and there is no `index.html` file

To what extent should this reference the other RFC? As we could talk about `default` entries vs the specific `index.html`....

text/0000-xor-urls/0000-xor-urls.md
```diff
@@ -0,0 +1,115 @@
...
+
+An specific MD entry key could be also supported (e.g. `__non_browsable`), that the owner of the MD can insert in the MD if the “browsable” feature should not be enabled. Although this cannot be really enforced but offered as a feature to be optionally supported by some clients like our SAFE browser.
```
joshuef: Would a client support `__non_browsable` but not support displaying the data in some format?

text/0000-xor-urls/0000-xor-urls.md
```diff
@@ -0,0 +1,115 @@
...
+
+This represents not only a restriction in some trivial use cases, like sharing a file by sharing its location and without publishing it under a public-name, but it also prevents users from being able to store linked-data on the network as these type of data cannot be linked unless it's all published at a public-name.
+
+## Detailed design
+
+### XOR address encoding
+The XOR address shall be encoded in the XOR-URL, [z-base32 encoding](http://philzimmermann.com/docs/human-oriented-base-32-encoding.txt) seems to be a good choice as it is case insensitive, as opposed to other case-sensitive encodings like base58btc or base64url, and it was designed to be easier for human use permuting the alphabet so that the easier characters are the ones that occur more frequently.
+
+### ImmutableData XOR-URL
+XOR-URLs for ImmutableData’s are the simplest ones since they don’t need any additional information to uniquely identify them on the network, as opposed to MutableData’s that also have a type tag. Therefore an ImmutableData XOR-URL can be simply defined as `safe://<encoded ImmutableData XOR addr>`.
```
joshuef: `safe://<encoded ImmutableData XOR addr>`.

Do we not need to include the encoding func as part of the url?

joshuef: Ah, you go into this below 👍

text/0000-xor-urls/0000-xor-urls.md
```diff
@@ -0,0 +1,115 @@
...
+
+- `<type-tag>`: the type tag value if the CID is referencing a MutableData. In the absence of this value, the CID will be assumed to be targetting an ImmutableData
+
+- `<path>`: the path of the file if the CID is referencing a MutableData, which can be accessed through the NFS/Files convention (or other emulations/conventions in the future) and resolve it
+
+- `<query>`: query arguments, to be used by the client app and not for retrieving the content
+
+- `<fragment>`: fragment of the content, to be used by the client app and not for retrieving the content
+
+The resolver function shall first attempt to decode the URL assuming it contains a `<cid>` part. If that step fails (either because it couldn’t decode the CID, or because there was no data at the decoded XOR address), it will do a fallback to assume it’s a public-name URL and it shall try to find the pulic-name MD (with the `sha3` hash of the string), just like any public-name URLs are resolved now by the resolver function.
```
joshuef: `The resolver function shall first attempt to decode the URL assuming it contains a `<cid>` part.`

This is slightly different to the PNS RFC atm. If this is the desired process, we should be sure to sync up the two. (more a note to self :P )

text/0000-xor-urls/0000-xor-urls.md
```diff
@@ -0,0 +1,115 @@
...
+
+NFS/Files container MutableData XOR-URL:
+`safe://hyfktcenm57js4bm3owhez9td9pi3t8bzk1crqp7mr5865c15ih3yxpz68w:15008/some/folder/index.html`
+
+## Drawbacks
+
+None identified as of yet.
+
+## Alternatives
+
```
joshuef: Another alternative is a straight up `safe-xor://` protocol for IDing these URIs... Maybe good to mention why we dont want that and why CID as id of XOR is enough.

text/0000-xor-urls/0000-xor-urls.md
bochaco: yeah, it makes sense

text/0000-xor-urls/0000-xor-urls.md
bochaco: Do you mean a link to the PNS RFC PR? I can do that yes, although as far as this RFC is concern I guess it's just if there was no index.html or default resolved from the NFS container found, right? or you are suggesting to have some more details here about the resolution too?

text/0000-xor-urls/0000-xor-urls.md
bochaco: I hesitated about having this part in this RFC since it's not really something related, and it in our case specific to a feature we can have in the browser I think. As you say, the client app can do anything with it really, so it's more about a convention and having our browser to follow it, but this cannot be enforced. In any case I'd agree to perhaps remove this from here and discuss it separately as a potential feature for the browser, and/or for our resolvable maps?

text/0000-xor-urls/0000-xor-urls.md
bochaco: right, yes this is the CID that is detailed below, which is actually part of the proposal, I guess if anyone proposes a different encoding it should also be discussed here.

text/0000-xor-urls/0000-xor-urls.md
bochaco: @joshuef I don't see a difference really, can you elaborate?
Perhaps the only difference is that I'm saying we should try to decode the CID and attempt to fetch it before falling back, rather than only decoding, is this what you see different?
The reason I'm proposing that is to not eclipse a public name which happens to be a valid CID, it will be eclipsed anyway if there exist content at the location encoded by the CID but at least not do it when there is no content. But this again is what's proposed here, so it's open for debate.

text/0000-xor-urls/0000-xor-urls.md
bochaco: oh yes, good point, I'll add that!

text/0000-xor-urls/0000-xor-urls.md
bochaco: I'm adding this explanation to that paragraph as it may help

text/0000-xor-urls/0000-xor-urls.md
joshuef: > it's just if there was no index.html or default resolved from the NFS container found, right?

right.

> or you are suggesting to have some more details here about the resolution too?

Nono, just aligning where makes sense 👍

text/0000-xor-urls/0000-xor-urls.md
joshuef: > I'd agree to perhaps remove this from here and discuss it separately as a potential feature for the browser, and/or for our resolvable maps?

Yeh I think it's more browser based. We can discuss in an issue or so. I don't think it's resolvable maps related as it's not really a feature /requirement of the of the data structs etc. (saying they MUST display the file contents when no resolution is found seems over prescriptive).

text/0000-xor-urls/0000-xor-urls.md
joshuef: Ah you're right. The part I'm talking about is discerning uris vs other graph entries when referencing from `default`. Sorry for confusion!

text/0000-xor-urls/0000-xor-urls.md
bochaco: agreed, I'll remove that part then

text/0000-xor-urls/0000-xor-urls.md
```diff
@@ -0,0 +1,115 @@
...
+<xor-uri>             := <immutable-data-uri> | <mutable-data-uri>
+
+<immutable-data-uri>  := <cid> <query-fragment>
+
+<mutable-data-uri>    := <cid> ':' <type-tag> ['+' <content-version>] <path-query-fragment>
+
+<path-query-fragment> := ['/' <path>] <query-fragment>
+
+<query-fragment>      := ['?' <query>] ['#' <fragment>]
+```
```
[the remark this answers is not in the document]
bochaco: I'm not sure what you mean by a `key` , are you by any chance thinking along the lines of what [they defined as human-readable CIDs](https://github.com/ipld/cid#human-readable-cids)? 
If so , we could consider having the browser to show a tooltip with the human-readable CID when hovering over a XOR URL in the address bar, if that happens to be useful, and/or have the browser to automatically convert a human-readable CID into the CID string when the user pasted/entered it in the address bar as it can be good for those who understand the format to fetch content at a any given xorname. I would say this can be all done but I doubt it should be part of the resolver itself but perhaps provided by the client app, e.g. in our case the browser.

As per having shorter CIDs like you suggested above @ustulation , it would be nice, but shortening it as you seem to imply would break the encoding of the CID string, so you'd need some type of encoding which can be shortened that way and still be able to get the (shortened) xor address out of it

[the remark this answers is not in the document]
ustulation: I meant MData Key from the key-value pair.

On shortning, IDK how it breaks it. So say in base x you get `hsh121difvoc..` etc., if you trimmed it to `hsh1` you could convert that back to repr in base 16 (which currently what the network understand) and use that portion. So it will give you are shorter XOR address like explained in the overall [comment here](https://github.com/maidsafe/rfcs/pull/281#pullrequestreview-174881487). But again, if that might be too much of a digression and could hijack comments directly (and more closely) related to this RFC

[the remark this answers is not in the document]
bochaco: For getting a specific MData key, then yes you could use the #fragment which is what it's the standard way of referring to a specific part of the document you are targeting with the URL, e.g. the WebID spec also makes use of the #fragment to refer to one of the subgraphs, so I wouldn't do that as part of the resolver's job, since the #fragment is usually used by the app consuming the content.

Shortening the xorname to 4 or 6 bytes wouldn't still give you a 4 or 6 bytes XOR-URL, remember the CID encodes some others bytes, so yes, it sounds like it can be done depending which encoding you use (e.g. CID in this case) you won't still be able to shorten it to 6 bytes but a few more, which could still be good I guess. 
Anyway I think there is a major concern with shortening it as you suggest, it could be breaking the immutability of the relationship between the XOR-URL and the target content, e.g. you give me a short XOR-URL, and I can fetch an immutable content since it can resolve to an unique xorname, until one day someone stores another piece of data (at a xorname which has the same initial bytes) which makes that XOR-URL un-resolvable as it doesn't resolve to an unique address, right?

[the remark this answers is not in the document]
ustulation: >  as it doesn't resolve to an unique address, right?
Ya that's true (it's same as git, as mentioned), but probably not worth the hassle as you say.

> For getting a specific MData key, then yes you could use the #fragment which is what it's the standard way of referring to a specific part of the document you are targeting with the URL
Would it be better if you exemplified it in the RFC ? Perhaps a few more examples to illustrate would give an easier understanding while reading the RFC. I would also propose to explain the deconstruction in examples if it's needed. Something like:
```
to do X: <abcd>:://<efgh>:<ijkl>?<mnop>#<qrst>
where abcd=cid
           efgh=blah
 ... etc.
```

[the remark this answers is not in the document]
bochaco: Agreed, I also think it will help to have some details of the decoding, I'll work on adding it @ustulation

text/0000-xor-urls/0000-xor-urls.md
```diff
@@ -0,0 +1,119 @@
...
+
+Note that it's proposed that the resolver should try to decode the CID and attempt to fetch the content before falling back to assume it's a public-name URL, rather than only decoding. The reason behind this is to not eclipse a public-name which happens to be a valid CID string. In such cases it could still be eclipsed if there exist content at the location encoded by the CID string, but at least not do it up front when there is no content stored at such location.
+
+The following are examples of what would become valid XOR-URLs as per this proposal:
+
+ImmutableData XOR-URL:
+`safe://hygjdkfty6m7ag3bckq7eqgeizbtjk915c3jbrcgtisad8iikbk4xws4jbpky`
+
+MutableData XOR-URL:
+`safe://hyfktce8j75yhmj1dbi1xw5wnb4m3zdydr7wpbzf1a16hc3sbxzu8a9hiqw:15000`
```
krishnaIndia: If we have a new data type which has a XOR name and tag type which behaves different to a MudatbleData, for example, consider the StructuredData is being revived and supported. How can we represent the same as a XOR url?
If my understanding is right it will also have a similar URL as a MutableData would have.

Potential options:
- The CID generation takes [registered codec](https://github.com/ipld/js-cid#new-cidversion-codec-multihash) as an input. But not sure whether custom codec DataTypes will be accepted as a part of the codec list to multicodec and cids repositories. Other options is to fork the repository which comes with maintenance overhead.
- Have different hashing techniques to differentiate Datatypes. When the URL is decoded the `webFetch` can do a switch based on the hashing techinque to resolve the corresponding datatype.

text/0000-xor-urls/0000-xor-urls.md
```diff
@@ -0,0 +1,119 @@
...
+```
+<safe-url>            := 'safe://' ( <xor-uri> | <public-name-uri> )
+
+<public-name-uri>     :=  [<service> '.'] <public-name> <path-query-fragment>
+
+<xor-uri>             := <immutable-data-uri> | <mutable-data-uri>
+
+<immutable-data-uri>  := <cid> <query-fragment>
+
+<mutable-data-uri>    := <cid> ':' <type-tag> ['+' <content-version>] <path-query-fragment>
```
krishnaIndia: While generating the digest of the multihash, we are using only the XOR name. And then we are adding tagType and version to the URL. For example, `safe://xor-address-of-md:tagType+version/sample.png`. This convention would make it difficult to parse IMO. 
Is it possible to have a json like representation,
```
{
  name: xor-name,//multibase encoded
  tageType: 15000,
  version: ''//string or number
}
```
This could be used to generate the CID like,
```
const getData = (address, tagType) => {
    return new Buffer(JSON.stringify({address, tagType}));
};
const encodedHash = mHash.encode(getData('somedata', 1000), 'sha1');
const cid = new CID(1, 'raw', encodedHash);
const encodedCID = cid.toBaseEncodedString('base32');
```
When decoded we have a defined structure to resolve.

text/0000-xor-urls/0000-xor-urls.md
```diff
@@ -0,0 +1,119 @@
...
+
+Another alternative specifically to the proposed XOR-URL format/encoding is to have a separate protocol, e.g. `safe-xor://`, for identifying these URIs. However there doesn't seem to be much benefits from doing this, and the only evident one is to avoid the scenario described above in drawback section when a public-name can be eclipsed by a XOR-URL, which it should be an unlikely scenario. On the other hand, having a single protocol as the standard for all type of URLs that resolve to content on the SAFE Network should be seen as a more homogeneous design.
+
+## Unresolved questions
+
+The current list of supported codecs in the [multicodec project](https://github.com/multiformats/multicodec) (which is part of the CID format that it's proposed to use) doesn't include the MIME-types that can be used for encoding in an ImmutableData XOR-URL. There is [a discussion](https://github.com/multiformats/multicodec/issues/4) which has been triggered to add such a support (see [PR sent](https://github.com/multiformats/multicodec/pull/84)) but it hasn't been approved yet.
```
krishnaIndia: Does the multicodec project allow to add multiple prefixes? which could be helpful to represent like [this](http://www.leanbackplayer.com/test/h5mt.html). I can think of scenario where an Video player app gets a XOR URI and needs to know of the mime type and also the codec. My knowledge in regards to codecs is limited, my use case could be wrong too 😉.

text/0000-xor-urls/0000-xor-urls.md
```diff
@@ -0,0 +1,119 @@
...
+ImmutableData XOR-URL:
+`safe://hygjdkfty6m7ag3bckq7eqgeizbtjk915c3jbrcgtisad8iikbk4xws4jbpky`
+
+MutableData XOR-URL:
+`safe://hyfktce8j75yhmj1dbi1xw5wnb4m3zdydr7wpbzf1a16hc3sbxzu8a9hiqw:15000`
+
+NFS/Files container MutableData XOR-URL:
+`safe://hyfktcenm57js4bm3owhez9td9pi3t8bzk1crqp7mr5865c15ih3yxpz68w:15008/some/folder/index.html`
+
+## Drawbacks
```
krishnaIndia: The XOR URI is limited only for public content sharing could be a limitation/drawback?

text/0000-xor-urls/0000-xor-urls.md
bochaco: Yes this is a good point, and I also thought a bit about it, and I was thinking more towards your first option. If the target content is an immutable data then the codec will likely be the one corresponding to the file itself, but if it's another type of 'structured' data like MutableData, then the codec can be our own custom set of types, which I think it should be possible since I see there are already a few custom ones in the table, like the ones for Ethereum: https://github.com/multiformats/multicodec/blob/master/table.csv#L416

text/0000-xor-urls/0000-xor-urls.md
bochaco: It could be, although I thought having the type tag and versions exposed in the URL could be better for the user. The type tag matches quite well if you think of an analogy with ports in the clearnet, and having the version available to the user to change and get a different version of the same content is also I thought could be good, and as mentioned in the RFC if you are willing to get the latest then you omit the version from the URL. Basically if you see the version in the URL explicitly could probably be better for the user as he/she knows which version is being fetched and even change it.

text/0000-xor-urls/0000-xor-urls.md
bochaco: It currently doesn't support it but I think that if this is a common practice out there by apps, i.e. to rely on several codecs for content, we should be able to propose that to the multicodec project and add such a support, it sounds to me it should be difficult at all to support it.

joshuef:
🐑 I've assigned myself as resident shepherd for this RFC. 🐑 

I'd love to get this merged as a proposed RFC. @ustulation currently we're pending approval of some changes you were after. Would you be so rad as to cast an eye over them, please?

PR #286, opened 2019-03-13
Refactoring the safe_client_libs and simplifying its API

bochaco:
After writting some Rust client apps using the safe_client_libs crates, I believe PR #213 can still be considered and still applicable, or at least it has some still valid considerations. I cannot re-open PR #213 (for some unknown rason) thus I'm just creating this new PR.

I'll be working on some updates to the original content of this RFC though, given I now have a bit more knowledge about the SCL lib.


nbaksalyar:
Looks great :+1: However, it should also be noted that the API is asynchronous in nature, so we might want to utilise the [upcoming async features of Rust](https://areweasyncyet.rs/).

PR #294, opened 2019-04-04
Regenerate the status document and add CI to fail if it needs updated

S-Coyle:
To regenerate the `RFCs-by-status.md` I ran:
```
./generate_by_status.sh > RFCs-by-status.md
```

I've also added a .travis.yml file which checks to see if any document status does not match the `RFCs-by-status.md`, and so forcing the author to add a commit to the PR to run the update. If CI fails it will provide a helpful message, as below:
image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,4 @@
+language: bash
+
+script:
+  - bash test_rfc_status.sh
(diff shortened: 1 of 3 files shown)


S-Coyle:
Currently testing - do not merge. Thanks

PR #295, opened 2019-04-05
Some corrections and enhancements to the text based on feedback received from the community so far

bochaco:


bochaco:
Yes, correct @joshuef , even if we had a decryption key provided with some mechanism, there is nothing we can do if it fails to decrypt. An straight answer could be, well we fail the request in such cases, but we also need to always remember that applications could corrupt data, e.g. by either adding non-conformant data entries, which we should ignore if we want to be able to still read the data that is valid. So basically, it's about doing the best we can to extract the data which follows the convention/spec/format our resolver conforms to even if it needs to coexist with other formats in the same data resource/object. This is how I see it at least.

bochaco:
This is obsolete now, and all changes needed as per last research are covered with PR #337

PR #329, opened 2019-07-04
RFC 55: Add details about the use of namespaces

lionel1704:


nbaksalyar:
This change no longer applies (the opposite is true: published and unpublished Immutable Data use different XOR namespaces).

PR #338, opened 2019-08-07
Update PNS -> NRS and add in more info

joshuef:
w/r/t new data types and versioning

### DIFF
--- a/text/0052-RDF-for-public-name-resolution/0052-RDF-for-public-name-resolution.md
+++ b/text/0052-RDF-for-public-name-resolution/0052-RDF-for-public-name-resolution.md
@@ -1,8 +1,8 @@
-# RDF for PublicName Resolution
+# RDF for the Public Name Resolution System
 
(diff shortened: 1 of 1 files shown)


b-zee:
Cool to see everything coming together!

A note: I see that the `v` query parameter (`?v=<x>`) will be used to address specific versions and examples of this are URLs like "`safe://thisxorurl#somewhere?v=1`". However, this specific kind of example is technically wrong. In the example the 'query' parameter **isn't** actually that because it is **after the segment component** and therefore a part of the segment.

In other words:
```
const url = require('url');
const url_arr = url.parse('safe://thisxorurl#somewhere?v=1');
// url_arr.hash == '#somewhere?v=1'
// url_arr.query == null
```

joshuef:
Closing this PR, as it's superseded by https://github.com/maidsafe/rfcs/pull/343 which contains all fixes in the PR-PRs.

PR #343, opened 2019-08-15
Updates to NRS nomenclature

joshuef:
And small fixes to previous PRs


vinitha-vasanthkumar:
@joshuef Not quite sure but is it supposed to be V0 to access all versions of the data?
image

PR #356, opened 2020-05-04
docs/rfc-0052: extend/alter nomenclature. add recursive resolution

dan-da:
First draft of some proposed changes to rfc-0052:

* introduce 'Top Names' nomenclature.
* introduce 'SAFE-URL` nomenclature.
* define 'Public Names' as SubNames + TopNames.  finally consistent with definition as "equivalent" to `host` in a standard URL.
* Add examples for usage of `Public Name`, `Top Name`, `Sub Name`.
* change URL 'host' reference to DNS 'domain', which is a better match
* Re-write Path Resolution section.
* Add a section on resolving recursive NRS-URLs

Note:  I considered naming `Top Name` `Global Name` instead.  This could reflect that it is a name in a truly global namespace.  I went with `Top` because it is shorter to type, and is an easy 1:1 mapping in people's minds with DNS top-level-domain. But if people prefer Global instead, it could be a decent choice.

### The issue: naming inconsistency in RFC-0052.

Let's examine this quote from RFC-0052:

> Here the URL terminology for host is equivalent to a SAFE Public Name. What are known as subdomains on the clearnet are referred to as SubNames.
>
> safe://&lt;subname>.&lt;publicname>?v=&lt;version>

Four interwoven issues arise with the original RFC-0052:

1. `Sub Name` `a` and `Public Name` `b` are each named, but the entire string `a.b` has no identifier.
2. In a standard clearnet URL containing a DNS domain name, the `host` field refers to the entire string `a.b`, NOT to a portion of it, ie `b`.  So `host` is NOT equivalent to `Public Name` as stated.
3. The `host` field of a URL may be interpreted as various things, eg: ipv4 addr, ipv6 addr, dns domain name, or another name resolution system, according to the url's scheme.  With "safe://" URL scheme, the `host` field can contain either XOR id or NRS name.  So, really the closest analogy is not URL `host`, but DNS `domain name`.
4. The entire `a.b` name is intended to be publically visible and resolvable.  Hence, it seems erroneous to call only the rightmost portion of it `public name`.  This gives the appearance that the subname portions are somehow not public, and is confusing.

In order to finalize my [pull-request](<https://github.com/maidsafe/safe-api/pull/529>) to SafeApi XorUrlEncoder (now SafeUrl), (1) and (4) need to be resolved somehow so we know what to name properties and methods and keep the code and RFC consistent. That is the real goal of these changes.

### DIFF
--- a/text/0052-RDF-for-public-name-resolution/0052-RDF-for-public-name-resolution.md
+++ b/text/0052-RDF-for-public-name-resolution/0052-RDF-for-public-name-resolution.md
@@ -14,7 +14,6 @@ This proposal looks to enhance the public name resolution system by using a reso
 
 ## Conventions
 - The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
-- XOR-URL refers to a url generated as part of the content addressable system for accessing `xorname` urls in the safe_browser. As described here: https://forum.safedev.org/t/xor-address-urls-xor-urls/1952
 - Data is presented as RDF, serialised in [JSON-LD](https://json-ld.org/) in the examples.
 - I'm using AppendOnlyData, `AOD`, and ImmutableData `ID` as shorthand.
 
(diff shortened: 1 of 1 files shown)


text/0052-RDF-for-public-name-resolution/0052-RDF-for-public-name-resolution.md
```diff
@@ -64,47 +67,117 @@ If so, it is resolved via XOR-URL.
...
 
+#### 3. Path resolution
+
+When resolving NRS-URL `link` to SAFE-URL `target`, the path of
+`link` is appended to the path of `target`.
+
+Example: An NRS-URL has been created with `Top Name` test:
+
+```
+safe://test  --> safe://<xor>/testdata
```
joshuef, line 139: It may well confuse the point, but a `v` would be needed here. Perhaps if we don't want to include it here, maybe just an asterisk to note it? (As re-reading we're not _that_ explicit about that requirement in this RFC).

text/0052-RDF-for-public-name-resolution/0052-RDF-for-public-name-resolution.md
```diff
@@ -154,7 +227,7 @@ Provides data to be shown at the public name.
 		  },
 		"@type": "ResolvableItem",
 		"@id": "safe://thisxorurl?v=2#email",
-		"target": "<target graph or safe url (xor or pubName); eg: 'email'>",
+		"target": "<target graph or safe url; eg: 'email'>",
```
joshuef, line 230: should this be `safe-url` as well?

text/0052-RDF-for-public-name-resolution/0052-RDF-for-public-name-resolution.md
dan-da: good point.  I will add it.

dan-da: Ok, I renamed the section to Path and Version resolution and gave rules/examples for both.  Please double-check it is correct/logical.

text/0052-RDF-for-public-name-resolution/0052-RDF-for-public-name-resolution.md
dan-da: yes, fixed.

dan-da:
I also just added a section `SAFE-URL Reserved Query String Parameters`, that reserves `v`, `safe` and `safe-*` for present or future use by the network.  any objections?

dan-da:
I just added another section `Fragment Resolution`.   It may be debatable if "safe nrs create" should allow creation of targets with fragment at all.  But it presently does allow, and further I reason that even if we prohibit it in rfc and code, other code/libs might still allow it, so it is a good idea to have clearly defined resolution rules.

PR #360, opened 2024-07-08
Initial proposal for short-term release process

jacderida:


0001-short-term-release-process/0001-release-process-short-term.md
```diff
@@ -0,0 +1,325 @@
...
+Like Gitflow, we would also have some temporary branches; these will be discussed in more detail
+later.
+
+### Merging
+
+Historically, our team has preferred `git rebase` to `git merge`; however, Gitflow is better
+supported by using `git merge`. Rebasing is fine before submitting upstream, but Gitflow involves
+merging commits between different branches. Due to the fact that rebasing rewrites the commit
+history, merging between branches can lead to commits that have the same content but different
+hashes. This can make merging more confusing than it needs to be and also cause the commit history
```
joshuef, line 103:
```suggestion
hashes. This can make subsequent merging, post-rebase, more confusing than it needs to be and also cause the commit history
```

0001-short-term-release-process/0001-release-process-short-term.md
```diff
@@ -0,0 +1,325 @@
...
+This section provides guidelines for possible dates for each cycle until launch.
+
+### Wave 1
+
+Release and deployment for next wave:
+* Stable release and production deploy: 2024-07-08
+
+### Wave 2
+
+Live for users: 2024-07-09 / 825 people / 11 week comp / comp ends: 2024-09-27
```
joshuef, line 50: > comp ends: 2024-09-27

what does `comp ends` mean here?

0001-short-term-release-process/0001-release-process-short-term.md
```diff
@@ -0,0 +1,325 @@
...
+    - When ready, perform a release and tag of `stable` by kicking off a workflow. *
+    - Delete the release branch.
+    - Deploy to `PROD-01`/`02` and drain *
+    - Announce to users
+
+We can accommodate hotfixes at any point during the cycle.
+
+#### Uncertainties
+
+* Not completely sure about the duration of each phase. Any opinions?
```
joshuef, line 145: I'd probably If we want to release to users for partial consumption (ie, say 5% of early adopters use the code), we'd probably want to go longer here to allow for this phase and subsequent assessment?

I'd say we want to be conservative here (and start so), with `alpha` or `beta` level code being released to the community before a final `rc`. (are all those required?)

0001-short-term-release-process/0001-release-process-short-term.md
```diff
@@ -0,0 +1,325 @@
...
+* We could deploy automatically, but do we need to coordinate with an announcement?
```
joshuef, line 146: I'd say only final release woudl need an announcement. Alpha/beta/rc could should go out as/when ready to keep the flow simpler.

Ideally it would allow for auto-updating if you've opted in to that release channel eg to keep the need for announcements there to a minimum. (Anything we want could/should just be added to the github release post?)

0001-short-term-release-process/0001-release-process-short-term.md
```diff
@@ -0,0 +1,325 @@
...
+crates must use Semantic Versioning. A binary is defined within a crate, and therefore, by default,
+it will also have a Semantic Version; however, it is possible to override the `--version` argument
+on the binary to provide something custom. Our releases currently produce eight binary artifacts. *
+It would be useful if we could refer to these collectively with a single version number and package,
+where the package name would reflect the version number. ** The `--version` argument would identify
+this version number, but also identify the individual component using its Semantic Version. *
+
+The proposal is to use `YYYY.MM.X.Y` as the collective version number. **
+
+We would use the collective version number for a single Github Release. The assets for the release
```
joshuef, line 167: Will the crates still have their individual version numbers? that's not clear to me here

0001-short-term-release-process/0001-release-process-short-term.md
```diff
@@ -0,0 +1,325 @@
...
+would be the combined binary packages for each platform. The changelog can also be nicely applied to
+this combined release.
+
+#### Uncertainties
+
+* Should `faucet` and `sn_auditor` be part of a package targeted at users?
```
joshuef, line 173: Hmm, yes. Both could be useful in local development.

0001-short-term-release-process/0001-release-process-short-term.md
```diff
@@ -0,0 +1,325 @@
...
+* Should Semantic Versions be dropped from user-facing elements?
+* What are the `X` and `Y` in the version number?
```
joshuef, line 175: x = release candidate status? (alpha/beta/rc/nothing=stable)
y = subversion of that release candidate status

?

0001-short-term-release-process/0001-release-process-short-term.md
```diff
@@ -0,0 +1,325 @@
...
+
+### Alpha Releases
```
joshuef, line 177: I would say this is a `hotfix` release candidate

0001-short-term-release-process/0001-release-process-short-term.md
```diff
@@ -0,0 +1,325 @@
...
+* If need be, any fix commits applied on the alpha branch should be cherry picked into `main`.
+* The experiment is over and the branch should be deleted.
+
+The branch is discarded because we don't want the alpha version bumps back in `main`. Crates were
+also not published. The Github Release will function as the historical record of the existence of
+the alpha release.
+
+#### Uncertainties
+
+* Would we need to use the `NETWORK_VERSION` compile-time variable for this?
```
joshuef, line 221: I think we should perhaps scrap this.

internal testnets won't cross over (unless we are not careful...but that would be mitigated by:)

We should work our way for the environment released into to have differing genesis keys, and that will in effect prevent this.

0001-short-term-release-process/0001-release-process-short-term.md
joshuef: ah, you have hotfixes below. 

As i note above, i see alphas more in terms of outward facing release readiness.

0001-short-term-release-process/0001-release-process-short-term.md
joshuef: I think in a released network we would want to have monthly/bimonthly release candidate and longer testing period to ensure stabliilty, 

So perhaps we start faster (and no alphas) and expand into this flow over the coming months?

0001-short-term-release-process/0001-release-process-short-term.md
jacderida: Ah sorry, the schedule here was actually based on a Slack post from Bux, and she put this phrase in each of the phases. I think it means "competition ends". I will clarify that with her and update.

0001-short-term-release-process/0001-release-process-short-term.md
jacderida: Just to be clear, there is no `beta` in this proposal. The `alpha` releases may not happen, and if they do, they are intended for very quick experiments. Otherwise, it just goes to release candidate.

When you say "start faster" are you suggesting we should have a shorter development cycle than two weeks? Still one week for the RC testing phase?

0001-short-term-release-process/0001-release-process-short-term.md
jacderida: In this proposal, we don't really have the notion of a persistent channel. If community members were to participate in either alpha or RC testing, they would just use specific version numbers. Just want to make sure we are both on the same page with that one.

0001-short-term-release-process/0001-release-process-short-term.md
jacderida: When it comes to crates, those are obligated to have semantic versions. If we're talking about binaries, I think it would be useful to keep the Semantic Version on each of the individual binaries; however, Shu is of the opinion that the Semantic Version should be removed from everything that's user facing. We need to make a decision on it.

0001-short-term-release-process/0001-release-process-short-term.md
jacderida: I'm not sure. This versioning scheme was suggested by Shu, so I'll let him fill in the blanks there.

0001-short-term-release-process/0001-release-process-short-term.md
jacderida: In this proposal, the alpha releases may or may not happen. I saw them as a way to have a very quick experiment as part of the current release cycle, *if we needed to do that*. It was based on conversations we had before about doing quick tests with the community on an isolated network. That's the function of the alpha in this proposal. If that doesn't come up, there would be no alpha component in the cycle. It would just go straight to `rc`. Again, this is for the short-term. That may not be what we'd do in the longer term.

If you are thinking that alpha is always part of the cycle, then I feel here we have a big point of contention that we'd need to make a decision on.

0001-short-term-release-process/0001-release-process-short-term.md
jacderida: OK, if we're confident there will be no interference, I will remove this.

0001-short-term-release-process/0001-release-process-short-term.md
```diff
@@ -0,0 +1,325 @@
...
+
+* Should `faucet` and `sn_auditor` be part of a package targeted at users?
+* Should Semantic Versions be dropped from user-facing elements?
+* What are the `X` and `Y` in the version number?
+
+### Alpha Releases
+
+An alpha release would accommodate a scenario in which we wanted to put out quick, experimental code
+to have community users test something on an isolated network. We can branch it off `main` and
+discard the branch when it's done. The branch is intended to have a very short duration. It will be
```
b-zee, line 181: Do we tag these releases? I would prefer if the git commits are referenced by something so they don't get lost.

jacderida: Of course, good catch. I'll get that added to the process.

0001-short-term-release-process/0001-release-process-short-term.md
jacderida: Thanks. I've removed this as an uncertainty.

Az80386:
Seems the development/release cycle is defined as a 3 week cycle, and written as part of 5 waves.

wave 2: development begins: 07-08 -> 07-22 (14 days) then cut happens
wave 3: development begins: 08-01 -> 08-15 (14 days) then cut happens
wave 4: development begins: 08-22 -> 09-05 (14 days) then cut happens
wave 5: development begins: 09-12 -> 09-26 (14 days) then cut happens

Am I misinterpreting this because development should always be ongoing on `main`, the way its written its as if development stops between 07-22 and 08-01 and so forth on the gaps between the 4 waves above.

Just throwing this question out there, other than the label `release candidate` vs `alpha`  coming from `main` branch carved as branch:  `YYYY-MM-DD-rc.1` or `YYYY-MM-DD-alpha.1` , its basically a build that is coming from `non-stable` branch going out to be tested, and later those changes get accepted later into `stable` and/or `main`, and if its the `release candidate` when adequate burn in happens with community (assuming this is true), (say one week) is then accepted into `stable`, and the stable release process kicks in.

- Does the team want this functionality and overhead here at this time?  Are we trying to have some of the community members trying to join alpha builds, as oppose to release candidates builds, and pick between one or both based on their limited resources and time in a very aggressive time line as is?
- At any point in time a `release candidate` or `alpha version` if given to the community doesn't necessarily have to be accepted and changes don't have to go back to `stable` or `main` respectively, and the team controls that decision.

>An alpha release would accommodate a scenario in which we wanted to put out quick, experimental code to have community users test something on an isolated network.

- Is the`release candidate` expected to go out to community members or only `alpha` builds?
- I believe the intention is for release candidate to go out to the community, but the top summary on the `release candidate` release section doesn't make a note off this compared to the alpha release summary, though its briefly mentioned within the `Process` section (kind of buried) as :

> Invite users to participate in testing the RC. They can use safenode-manager to obtain the rc nodes and safeup for rc clients.

- Just to confirm, is a `release candidate` expected to run as an isolated network when asking the community to test it?

0001-short-term-release-process/0001-release-process-short-term.md
```diff
@@ -0,0 +1,322 @@
...
+* Release candidate branch cut: 2024-09-05
+* Stable release and production deploy for wave 5: 2024-09-12
+
+### Wave 5
+
+Live for users: 2024-09-13 / 1000 people / 3 week competition / competition ends: 2024-09-27
+
+Three week development/release cycle:
+* Development begins 2024-09-12
+* Release candidate branch cut: 2024-09-26
```
Az80386, line 81: Seems the development/release cycle is defined as a 3 week cycle, and written as part of 5 waves.

wave 2: development begins: 07-08 -> 07-22 (14 days) then cut happens
wave 3: development begins: 08-01 -> 08-15 (14 days) then cut happens
wave 4: development begins: 08-22 -> 09-05 (14 days) then cut happens
wave 5: development begins: 09-12 -> 09-26 (14 days) then cut happens

Am I misinterpreting this because development should always be ongoing on main, the way its written its as if development stops between 07-22 and 08-01 and so forth on the gaps between the 4 waves above.

0001-short-term-release-process/0001-release-process-short-term.md
```diff
@@ -0,0 +1,322 @@
...
+### Release Candidates
+
+A release candidate (RC) is the binary that's intended to be released as a stable version. The set
+of features and fixes in the RC is what's included on `main` in the current cycle, i.e., between now
+and the last stable release. After about two weeks of development we should produce the RC for
+testing in the staging environment.
+
+An owner should be designated to the process, and it would be useful for this to cycle through
+everyone in the team. Once the RC branch is started, we won't accept new features on it, only fixes.
+Feature development can continue on `main`.
```
Az80386, line 227: - Is the`release candidate` expected to go out to community members or only `alpha` builds?
- I believe the intention is for release candidate to go out to the community, but the top summary on the `release candidate` release section doesn't make a note off this compared to the alpha release summary, though its briefly mentioned within the `Process` section (kind of buried) as :
>Invite users to participate in testing the RC. They can use safenode-manager to obtain the rc nodes and safeup for rc clients.


- Just to confirm, is a `release candidate` expected to run as an isolated network when asking the community to test it?

0001-short-term-release-process/0001-release-process-short-term.md
Az80386: The X and Y in my original diagram was defined as:

X = The release cycle itself within a given month, 1 being the 1st release cycle, 2 being the second (say 2 release cycles within a month) (different than the 3 week cycle proposal above)
Y = Just a counter that starts at 1 for a given release cycle and if more builds are made (due to hot fixes), it increments the version here for the collective build package/artifacts as is.  It would restart at 1 for say the second release cycle of the month here.

0001-short-term-release-process/0001-release-process-short-term.md
Az80386: My vote on the binaries is to have 2 separate flags, `--version` that shows the collective version as is say `YYYY.MM.X.Y` (easy for the consumer to relate all their artifacts are from the same bundle), and `--protocol-version` that reflects the `sn_networking` crate's semantic versioning (the actual label for the `--protocol-version` can be changed if desired).

0001-short-term-release-process/0001-release-process-short-term.md
joshuef: Ah, no i did not mean less than two weeks.

I mean: I think an ideal cadence in a live and functioning network will be many many weeks, maybe months. To be sure. (Bitcoin/eth is like one/6 months eg).

That is obviously quite long for us now. So perhaps we start with shorter cycles but increase them as we go?

(which would in turn offer time for alpha/beta releases with better burn in testing)

0001-short-term-release-process/0001-release-process-short-term.md
joshuef: Yeh, that makes sense.

The "channel" concept is more like "what kind of releases would you like to auto-update to". With no auto updating, there is no "channel", people just run what they want manually.

0001-short-term-release-process/0001-release-process-short-term.md
joshuef: Is it then something like 

`bundle-version` for a release bundle and `version` for the actual version of a crate and `protocol-version` for the version of the protocol?

Seems like we can have it all, if we're clear about it. 

`bundle version` seems fine and simple as the default user facing output?

0001-short-term-release-process/0001-release-process-short-term.md
joshuef: Okay, that seems fine to me.

0001-short-term-release-process/0001-release-process-short-term.md
joshuef: Okay, so `alpha`/`beta` are signifiers of testing code. 

Whereas things in the release cycle are always `rcs` as they are always intended to be going to production. There is nothing "unknown" about them.

That seems a fine and good distinction to make. I'm happy for that 👍

0001-short-term-release-process/0001-release-process-short-term.md
joshuef: I think `rc` goes out to community and would be initial integration into existing networks.

`alpha/beta` as @jacderida explained above is about experiments and separate networks (which may or may not go out to community. I think these correspond to what would be `DEV-01` eg testnets in the naming schemes

joshuef:
> Does the team want this functionality and overhead here at this time? 

I think we don't necessarily have to leverage it, but if it's clear from this proposal we'll at least be on the same page. I agree with chris' proposal here as I understand it, so we have the naming to reflect stability / testing status. 

As when those might be put out to the community would depend on what we need to test/how eg.

0001-short-term-release-process/0001-release-process-short-term.md
jacderida: Ah yeah, ok! I agree: I was initially thinking like 6 weeks. That then led me to narrowing the scope for this proposal to just cover the short term, which is for getting us to the initial release. Our process will probably be significantly different for post-release. I'm going to just go with what's stated in the proposal then, which is a three week cycle with two phases: two weeks for development, one week for RC testing, then deploy.

0001-short-term-release-process/0001-release-process-short-term.md
jacderida: OK cool. Again, this is something that will probably be different for post-release, but for short term, it sounds like we are in agreement that there are no channels.

0001-short-term-release-process/0001-release-process-short-term.md
jacderida: Thanks for the input on this.

Sorry Josh, but your last statement introduces a bit of uncertainty here for me. I interpreted your first statement as this:
`--version`: outputs the Semantic Version (crate version)
`--bundle-version`: outputs the packaged `YYYY.MM.DD.X.Y` version
`--protocol-version`: outputs the network protocol version

However, you then seem to suggest that `--version` would output the bundle version?

I honestly don't see why we couldn't just have the bundled version and the Semantic Version on the `--version` output? I want to keep the Semantic Version in because I think it's useful to be able to distinguish the individual components and also because `safeup` and `safenode-manager` works on the Semantic Versions. I honestly don't believe our time would be well spent just now on that refactor. I don't see why it's so unreasonable for the user to see the Semantic Version.

I don't want to keep going round in circles on this though. Shu has made his argument and I've made mine. Would you be able to clarify your point on the bundled version and help us reach a conclusion here please?

0001-short-term-release-process/0001-release-process-short-term.md
jacderida: OK cool, thanks.

jacderida:
> Just throwing this general question out there, other than the label `release candidate` vs `alpha` coming from `main` branch carved as branch: `YYYY-MM-DD-rc.1` or `YYYY-MM-DD-alpha.1` , its basically a build that is coming from `non-stable` branch going out to be tested, and later those changes get accepted later into `stable` and/or `main`, and if its the `release candidate` when adequate burn in happens with community (assuming this is true), (say one week) is then accepted into `stable`, and the stable release process kicks in.
> 
>     * Does the team want this functionality and overhead here at this time?  Are we trying to have some of the community members trying to join alpha builds, as oppose to release candidates builds, and pick between one or both based on their limited resources and time in a very aggressive time line as is?
> 
>     * At any point in time a `release candidate` or `alpha version` if given to the community doesn't necessarily have to be accepted and changes don't have to go back to `stable` or `main` respectively, and the team controls that decision.

I actually do agree with you that it could potentially be a bit confusing to have the alpha releases in three week cycles, but I think given what Josh has said, I'm going to just leave it in the proposal. I think it's unlikely that it's something that we will actually make use of, but it's there just in case. Thanks for the input on this one.

0001-short-term-release-process/0001-release-process-short-term.md
jacderida: The release candidate was definitely intended to go to the community. I will push a change to make that more clear in the first paragraph.

In my mind, I was thinking about the RC running on an isolated network, to be tested in isolation.

I would be interested to know what you think here Shu. If we are pushing it to the real network, would we add nodes to the existing network, or upgrade some current nodes? How would we verify our changes were working?

Should the three of us perhaps have a call to discuss this one? Just keen to reach a consensus and move the process forward as soon as possible.

0001-short-term-release-process/0001-release-process-short-term.md
Az80386: Adding in that for monitoring and grafana dashboards, having multiple versions in a `--version` output would end up storing multiple versions in a single string within a single column in the dashboards since there is no way to separate these out, including the output that comes from `safenode-manager status --json` for version.

My vote is to keep all 3 as separate parameters, so we can get all 3 version into separate columns on the monitoring as well.

jacderida: The decision has to be balanced against the refactoring that would need to be done for `safeup` and `safenode-manager` to use the different version number formatting. I think it's too much of a glaring inconsistency to have those tools use Semantic Versions when those versions are not returned with `--version` in the binaries.

So, I still vote against this, even with the Dashboard impact. Need a decision from Josh here.

joshuef: seems like we can have 

```--package-version: outputs the packaged YYYY.MM.X.Y version
--crate-version: outputs the Semantic Version (crate version)
--protocol-version: outputs the network protocol version``` to get the dashboard on track.

And then the only question is what output do we get/want from `--version`? And is there any reason not to just output it all? (which, if all things can be gathered separately, i don't think there is?)

jacderida: I think the above suggestion is what we should go with.

0001-short-term-release-process/0001-release-process-short-term.md
jacderida: In a call we agreed that the RC would be tested in isolation.

joshuef (approved):

